PR #20, opened 2021-11-11
Add xPOP deployment

ecmendenhall:
- Add xPOP deployment script:
    - Mint 1MM xPOP
    - Transfer ownership to multisig
- Add multisig addresses for Polygon + Arbitrum
- Update tag for setup deployment from `"LBP"` to `"setup"`


ecmendenhall:
Note: Should we have hardhat deployments for the `localhost` network checked in? Every local `hardhat node` run updates the block hash in each deployment.

anthonymartin (approved):

PR #26, opened 2021-11-17
Landing Page

mefuru:
Closes #642


RedVeil:
The page looks great and adjusts well on all the screen sizes that i saw. I am a bit wary though if using inline styling with pixel to adjust for a ton of screen sizes. It works here as their is not too much content but for other pages in the future we should rely more on tailwindcss and do some research in order to find the most popular screen sizes and optimize for them.

RedVeil (approved):

mefuru:
> The page looks great and adjusts well on all the screen sizes that i saw. I am a bit wary though if using inline styling with pixel to adjust for a ton of screen sizes. It works here as their is not too much content but for other pages in the future we should rely more on tailwindcss and do some research in order to find the most popular screen sizes and optimize for them.

Agreed. The designs are specified to the px so unsure how best to adhere to the designs as closely as possible and to rely on tailwind css. We could create tailwind design specifications for breakpoints and commonly used spacing and adhere to them while designing.

PR #27, opened 2021-11-18
Refactor/extract demo data

RedVeil:
Refactored the deployment for hardhat. Demo data gets only created on network hardhat.
namedAccounts contains all addresses now.


RedVeil:
DEPENDS ON hysi-page

anthonymartin (approved):

PR #33, opened 2021-11-26
Feature/network switcher

bhanusanghi:
IMPORTANT - Includes Major refactor to the current code base. 

The app was architected with Static Defined Pools and Pool Names.
This cannot work when we switch chains and the pools differ on the different chains.

Now Pools are loaded and populated dynamically. The list of staking pools is still maintained on FE codebase and I suggest we move it to a simple web2 API for scalability.
Rewards and Staking Pages have been refactored to comply with the new Architecture.


bhanusanghi:
Currently knows issues -

When we switch a chain the contract objects that existed earlier still persist for a minute timeframe and an error is shown on the UI for the same.

Pponwili1998:
#33

PR #40, opened 2021-11-30
Feature/web3 modal

Amirjab21:
I've replaced web3react with our own version of web3react that uses web3modal under the hood.

Annoyingly, metamask and web3modal have a well known bug which I can't appear to figure out a hack for.

The bug is that once you connect with metamask, it then keeps metamask in cache and won't let you use any other wallets unless you open an incognito window.

Here are discussions about this bug:

https://github.com/Web3Modal/web3modal/issues/183
https://ethereum.stackexchange.com/questions/83914/how-to-disconnect-metamask-wallet-using-web3modal?rq=1


anthonymartin:
@Amirjab21 can you take a look at the conflicts? looks like there are a handful here

Amirjab21:
Will do

I'm having to recreate parts of web3Modal as there are tons of issues that haven't been resolved so I'll do that first and ensure its fully functional

Pponwili1998:
1e423aa80eab27760df8bd8623e1a297bc1dea03

PR #77, opened 2021-12-20
Feature/fork cvx locker

RedVeil:


packages/hardhat/contracts/core/dao/PopLocker.sol
```diff
@@ -0,0 +1,1694 @@
...
+  }
+
+  // Claim all pending rewards
+  function getReward(address _account) public nonReentrant updateReward(_account) {
+    for (uint256 i; i < rewardTokens.length; i++) {
+      address _rewardsToken = rewardTokens[i];
+      uint256 reward = rewards[_account][_rewardsToken];
+      if (reward > 0) {
+        rewards[_account][_rewardsToken] = 0;
+        uint256 payout = reward / uint256(10);
```
RedVeil, line 1613: My bad, this should use .div() and .mul() since its not using solidity 8.0

PR #88, opened 2021-12-23
Fix flaky RewardsEscrow test

ecmendenhall:
Fix a couple flaky time related assertions in `RewardsEscrow` tests.

- Add `expectBigNumberCloseTo` test helper


RedVeil (approved):

RedVeil:
I quite like this for amounts that get muliplied or divded weirdly, we should just be careful to not use it too much.

PR #457, opened 2022-02-16
Added reverse lookup mapping to contractRegistry and a view function to retrieve it

bhanusanghi:
Add the relevant lookup mapping,  view function to retrieve it and tests to existing test suite to make sure all things work perfectly and bug free


anthonymartin (changes requested):

packages/hardhat/contracts/core/utils/ContractRegistry.sol
```diff
@@ -48,6 +54,10 @@ contract ContractRegistry is IContractRegistry {
     return contracts[_name].contractAddress;
   }
 
+  function getContractFromAddress(address _contractAddress) external view override returns (bytes32) {
```
anthonymartin, line 57: shouldn't this say getContractNameFromAddress ?

ecmendenhall: Or perhaps `getContractIdFromAddress`, since these are keccak hashes of the name + id.

RedVeil (approved):

PR #483, opened 2022-02-23
EIP-4626 Vault

ecmendenhall:
Add an [EIP-4626](https://eips.ethereum.org/EIPS/eip-4626) compliant Yearn vault wrapper contract. EIP-4626 is a proposed standard for yield vaults.

### How it works
The Popcorn `Vault` is designed to wrap a Yearn vault. We do so using Yearn's official `AffiliateToken` contract, available [here](https://github.com/yearn/brownie-wrapper-mix).

**EIP-4626 functions**
Our earlier iteration of the `Vault` had a `deposit` function to exchange assets for vault shares, and a `withdraw` function that redeemed vault shares for assets. Note how `deposit` deals with assets and `withdraw` deals with shares? 4626 introduces new `mint` and `redeem` functions that operate in terms of _shares_, and `deposit` and `withdraw` functions that operate on the underlying asset. That means the old function we called `withdraw` is now called `redeem`. 

4626 also adds some helper functions to preview the effects of `mint`/`deposit` and `withdraw`/`redeem`. Note that `withdraw` takes into account a 50bps withdrawal fee.

See the EIP [spec](https://eips.ethereum.org/EIPS/eip-4626) for all the details.

**Block Locks**
The `blockLocked` modifier, inspired by [Badger](https://github.com/Badger-Finance/badger-system/blob/beadfc54042ec57ae51534c252aa613bb630f60d/contracts/badger-sett/SettV4.sol#L117), prevents deposits/withdrawals/transfers within the same block. This is intended as a defense against flash loan attacks.

**Fees**
The Popcorn `Vault` takes both a "performance" and "management" fee on the underlying TVL. Fee calculations are based on total assets (`assetsCheckpoint`) and vault share value "high water mark" (`vaultShareHWM`) checkpoints. Fees are harvested and checkpoints are saved on every deposit/mint/withdraw/redeem interaction with the contract, as well as external calls to the `takeFees` function. 

If vault share value has increased above `vaultShareHWM` during a fee period, the vault collects a "performance fee" in shares equal to 20% of the returns.

The "management fee" is assessed using the checkpointed TVL at the start and end of the fee period. This fee is annualized per minute, using 525,600 minutes/year. Specifically, the management fee is assessed as:

```
(TVL at last checkpoint + Current TVL) * (fee period in minutes) * (management fee in BPS) / (2 * minutes per year)
```
This is like using the trapezoid method to evaluate a definite integral.

(**Update:** I think management fees are incorrect and can be simpler, see my comment [here](https://github.com/popcorndao/sweet-caramel/pull/483#discussion_r818871570)).

Fees are collected by issuing new shares to the vault, diluting value of existing shares.

### Yearn dependencies and Solidity `^0.6.12`
The Yearn dependencies all target Solidity `^0.6.12`, which require a few changes in order to support them:
- Using the aliased `openzeppelin-v3` package (version `3.4.2`) to make sure we're using the same version as our dependencies.
- Ensuring we use `SafeMath` for all our math operations in the `Pool`.
- Using `ACLAuthV6` and `ContractRegistryAccessV6`, versions of `ACLAuth` and `ContractRegistryAccess` with `public` constructors.

We have made a few modifications to these dependencies that we should consider carefully:
- Modification to `AffiliateToken`: change `pricePerShare` visibility from `external` to `public`, so we can call it internally in the `Vault` contract.
- Modification to `BaseWrapper`: make `totalAssets` a  `virtual` function and override it in `Vault`. Without this change, the `Vault` reports the TVL of the wrapped Yearn vault as its own. 

### Other changes:
- Add `IEIP4626` interface. Inheriting from this interface in `Vault.sol` ensures that we conform to the same interface defined in the EIP spec, and throws compiler errors when we don't.
- Add Yearn dependencies: `AffiliateToken`, `BaseWrapper` and `BaseStrategy`.
- Add `ACLAuthV6` and `ContractRegistryAccessV6`. These are versions of our existing `ACLAuth` and `ContractRegistryAccess` modified for Solidity `^0.6.12`. The only change here is declaring their constructors `public`.
- Import and update `PoolBlockLockHelper.sol`, a test helper contract for testing the "block lock" on the `Vault`. This prevents "block locked" functions from being called in the same block. 
- Add vault asset addresses to `getNamedAccounts`
- Add fork tests for target assets: `crv3Crypto`, `crvAave`, `crvComp`, `crvCvxCrv`, `crvSteth`, and `crvIbBtc`. Each fork test runs a scenario where two participants make multiple deposit/mint withdraw/redeem calls, with two rounds of simulated yield. 
- Simplify the `Faucet.sol` test helper: We have yet to find a stable abstraction that works for all Curve pools. The earlier version of the `Faucet` worked for _metapools_, but not for lending pools like `crvComp` or V2 pools like `crv3Crypto`. Instead of trying to abstract over any Curve pool, it is now boring and explicit: one method per LP token.
- Add a zero-address check in `ContractRegistryAccess`.


ecmendenhall:
> @RedVeil @ecmendenhall @yaoandrew - have any idea why fork tests are failing?

Yes, I'm pretty sure `process.env.RPC_URL` is unset in the CI environment again. Can you check on it in the CI settings?

packages/hardhat/contracts/core/defi/vault/Vault.sol
```diff
@@ -0,0 +1,634 @@
...
+  }
+
+  /* ========== RESTRICTED FUNCTIONS ========== */
+
+  /**
+   * @notice Set fees in BPS. Caller must have DAO_ROLE from ACLRegistry.
+   * @param newFees New `feesInBps`.
+   * @dev Value is in basis points, i.e. 100 BPS = 1%
+   */
+  function setFees(FeesInBps memory newFees) external onlyRole(DAO_ROLE) {
```
anthonymartin, line 401: DAO_ROLE should be changed to VAULT_MANAGER_ROLE

packages/hardhat/contracts/core/defi/vault/VaultFeeController.sol
```diff
@@ -0,0 +1,66 @@
+// SPDX-License-Identifier: GPL-3.0
+// Docgen-SOLC: 0.8.0
+pragma solidity ^0.8.0;
+
+import "../../utils/ACLAuth.sol";
+import "../../utils/ContractRegistryAccess.sol";
+
+contract VaultFeeController is ACLAuth, ContractRegistryAccess {
+  struct FeesInBps {
+    uint256 withdrawal;
```
anthonymartin, line 10: Should add deposit fee  to future proof

packages/hardhat/test/fork/rewards-distributon.test.ts
```diff
@@ -23,15 +23,15 @@ describe("rewards distribution test", () => {
       });
     });
 
-    it("should allow rewards manager to approve multiple times", async () => {
+    it.skip("should allow rewards manager to approve multiple times", async () => {
```
anthonymartin, line 26: Uh oh, we shouldn't be skipping these probably

packages/hardhat/contracts/core/defi/vault/Vault.sol
```diff
@@ -0,0 +1,728 @@
...
+    uint256 withdrawalShares = shares - withdrawFeeInShares;
+    // what is the fee in base asset?
+    uint256 fee = _shareValue(withdrawFeeInShares);
+    // what is the withdrawal amount in base asset after subtracting withdrawal fee?
+    uint256 withdrawal = _shareValue(withdrawalShares);
+    address rewardsManager = _getContract(REWARDS_MANAGER_ID);
+
+    _burn(owner, shares);
+    _withdraw(address(this), receiver, withdrawal, true);
+    _withdraw(address(this), rewardsManager, fee, true);
```
anthonymartin, line 586: this function should only be called if fee > 0

packages/hardhat/contracts/core/defi/vault/AffiliateToken.sol
```diff
@@ -62,21 +61,21 @@ contract AffiliateToken is ERC20, BaseWrapper {
...
+      return (totalVaultBalance(address(this)) * numShares) / totalShares;
     } else {
       return numShares;
     }
   }
 
-  function pricePerShare() external view returns (uint256) {
-    return totalVaultBalance(address(this)).mul(10**uint256(decimals())).div(totalSupply());
+  function pricePerShare() public view returns (uint256) {
+    return (totalVaultBalance(address(this)) * (10**uint256(decimals()))) / totalSupply();
```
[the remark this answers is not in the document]
RedVeil: check where pricePerShare is used and protect these cases from errors cause by totalSupply == 0

packages/hardhat/contracts/core/defi/vault/Vault.sol
```diff
@@ -0,0 +1,634 @@
...
+
+  /* ========== INTERNAL FUNCTIONS ========== */
+
+  /**
+   * @notice Internal helper for deposits.
+   */
+  function _depositFor(uint256 amount, address receiver) internal returns (uint256) {
+    require(amount <= token.balanceOf(msg.sender), "Insufficient balance");
+    require(receiver != address(0), "Invalid receiver");
+    _lockForBlock(msg.sender);
```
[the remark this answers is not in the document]
RedVeil: -- For now fine but if a contract wants to interact with it it can do so only once per block

packages/hardhat/contracts/core/defi/vault/AffiliateToken.sol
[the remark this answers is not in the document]
RedVeil: totalSupply is called three times in the contract. The only time totalSupply is not handled is in this view function. PricePerShare is only called externally so i think its ok if this breaks before anyone puts money in.

anthonymartin:
@RedVeil looks like some of the recent changes are causing all the CI stages to fail

packages/hardhat/contracts/core/defi/vault/AffiliateToken.sol
```diff
@@ -38,7 +38,6 @@ contract AffiliateToken is ERC20, BaseWrapper {
       abi.encode(DOMAIN_TYPEHASH, keccak256(bytes(name)), keccak256(bytes("1")), _getChainId(), address(this))
     );
     affiliate = msg.sender;
-    _setupDecimals(uint8(ERC20(address(token)).decimals()));
```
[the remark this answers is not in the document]
RedVeil: This is now resolved https://github.com/popcorndao/sweet-caramel/blob/d3512f30c29eb2cfbaaf9688d900b34e7d9a7846/packages/hardhat/contracts/core/defi/vault/AffiliateToken.sol#L43

packages/hardhat/contracts/core/interfaces/IPool.sol
```diff
@@ -6,7 +6,9 @@ pragma solidity ^0.8.0;
 interface IPool {
```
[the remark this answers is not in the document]
RedVeil: IPool was not used anywhere. Updating the name to IVault shadowed another IVault used in the LBP-Manager. Therefore i deleted IPool / IVault

packages/hardhat/contracts/core/defi/vault/Vault.sol
RedVeil: This was reworked in https://github.com/popcorndao/sweet-caramel/pull/483/commits/9f49418e92df101af835193a2436d08dfc057c4c

packages/hardhat/contracts/core/defi/vault/Vault.sol
RedVeil: This would be the only thing in the whole repo that uses VAULT_MANAGER_ROLE. Should we change the other management functions to use the VAULT_MANAGER_ROLE aswell? Otherwise id stick with the DAO_ROLE here

packages/hardhat/contracts/core/defi/vault/Vault.sol
```diff
@@ -0,0 +1,746 @@
...
+   * @param owner Owner of burned vault shares.
+   * @return Quantity of vault shares burned in exchange for underlying `asset` tokens.
+   */
+  function withdraw(
+    uint256 assets,
+    address receiver,
+    address owner
+  ) external override nonReentrant blockLocked onlyApprovedContractOrEOA returns (uint256) {
+    uint256 shares = previewWithdraw(assets);
+    require(owner == msg.sender, "sender must be owner");
```
RedVeil, line 424: This is not really ERC-4626 compliant but the openZeppelin version doesnt really let us play around with manual allowance setting.

This is how its handled in solmate:
```
if (msg.sender != owner) {
            uint256 allowed = allowance[owner][msg.sender]; // Saves gas for limited approvals.

            if (allowed != type(uint256).max) allowance[owner][msg.sender] = allowed - shares;
        }
```

Problem is that in openZeppelin these things are private so we cant even override it. So we might need to add our own allowance and approval directly in the Vault or deal with this limitation

packages/hardhat/contracts/core/defi/vault/Vault.sol
```diff
@@ -0,0 +1,746 @@
...
+  }
+
+  /**
+   * @notice Transfer accrued fees to rewards manager contract. Caller must have DAO_ROLE from ACLRegistry.
+   */
+  function withdrawAccruedFees() external keeperIncentive(contractName, 0) {
+    uint256 balance = balanceOf(address(this));
+    uint256 withdrawal = _shareValue(balance);
+    _burn(address(this), balance);
+    _withdraw(address(this), _feeController().feeRecipient(), withdrawal, true);
```
RedVeil, line 500: Note: We send funds now to the feeRecipient which is set on the feeController. We must make sure that this is not address(0) before withdrawing fees

packages/hardhat/contracts/core/defi/vault/Vault.sol
```diff
@@ -0,0 +1,746 @@
...
+    return shares;
+  }
+
+  /**
+   * @notice Internal helper for deposit and staking.
+   */
+  function _depositAndStakeFor(uint256 assets, address receiver) internal returns (uint256) {
+    require(staking != address(0), "staking is disabled");
+    uint256 mintedShares = _depositFor(assets, address(this));
+    IStaking(staking).stakeFor(mintedShares, receiver);
```
RedVeil, line 553: This uses Staking now. This means a user deposits for the Vault and the Vault stakes for the user. When a user unstakes they own the share token and can burn them again for underlying.

packages/hardhat/contracts/core/defi/vault/Vault.sol
```diff
@@ -0,0 +1,746 @@
...
+
+    // subtract the withdrawal fee from shares being withdrawn
+    uint256 withdrawalShares = shares - withdrawFeeInShares;
+
+    // what is the withdrawal amount in base asset after subtracting withdrawal fee?
+    uint256 withdrawal = _shareValue(withdrawalShares);
+
+    _burn(owner, shares);
+    _withdraw(address(this), receiver, withdrawal, true);
+    if (withdrawFeeInShares > 0) {
```
RedVeil, line 593: Is no conditional if we have any fees

packages/hardhat/contracts/core/defi/vault/Vault.sol
```diff
@@ -0,0 +1,746 @@
...
+    uint256 mintedShares = _depositFor(assets, address(this));
+    IStaking(staking).stakeFor(mintedShares, receiver);
+    return mintedShares;
+  }
+
+  /**
+   * @notice Internal helper for minting and staking.
+   */
+  function _mintAndStakeFor(uint256 shares, address receiver) internal returns (uint256) {
+    require(staking != address(0), "staking is disabled");
```
RedVeil, line 561: Only stakes if staking is set. Can be used to disable staking down the road

packages/hardhat/contracts/core/defi/vault/Vault.sol
```diff
@@ -0,0 +1,775 @@
...
+    address receiver,
+    address owner
+  ) internal returns (uint256) {
+    require(shares <= balanceOf(owner), "Insufficient vault share balance");
+    require(receiver != address(0), "Invalid receiver");
+
+    _lockForBlock(msg.sender);
+    _takeFees();
+
+    // calculate the withdrawal fee using the number of shares
```
anthonymartin, line 611: should use previewRedeem instead of the next few lines

packages/hardhat/contracts/core/defi/vault/Vault.sol
```diff
@@ -0,0 +1,813 @@
...
+    uint256 management;
+    uint256 performance;
+  }
+
+  address public staking;
+
+  bytes32 public contractName;
+
+  uint256 constant MINUTES_PER_YEAR = 525_600;
+  bytes32 constant REWARDS_MANAGER_ID = keccak256("RewardsManager");
```
anthonymartin, line 45:
```suggestion
```

packages/hardhat/contracts/core/defi/vault/Vault.sol
```diff
@@ -0,0 +1,813 @@
...
+  }
+
+  /**
+   * @return Maximum amount of vault shares that may be minted to given address.
+   */
+  function maxMint(address) external view override returns (uint256) {
+    return type(uint256).max;
+  }
+
+  /**
```
anthonymartin, line 286: @ecmendenhall @RedVeil 
 should  maxDeposit similarly use `previewDeposit`?

packages/hardhat/contracts/core/defi/vault/Vault.sol
```diff
@@ -0,0 +1,813 @@
...
+   * @notice
+   *  Used to update the yearn registry.
+   * @param _registry The new _registry address.
+   */
+  function setRegistry(address _registry) external onlyRole(DAO_ROLE) {
+    _setRegistry(_registry);
+  }
+
+  /**
+   * @notice Transfer accrued fees to rewards manager contract. Caller must have DAO_ROLE from ACLRegistry.
```
anthonymartin, line 606:
```suggestion
   * @notice Transfer accrued fees to rewards manager contract. Caller must have DAO_ROLE from ACLRegistry.
   * @dev we send funds now to the feeRecipient which is set on the feeController. We must make sure that this is not address(0) before withdrawing fees
```

RedVeil (approved):

packages/hardhat/contracts/core/defi/vault/Vault.sol
```diff
@@ -0,0 +1,813 @@
...
+
+  /**
+   * @return Total amount of underlying `asset` token managed by vault.
+   * @dev This function overrides the parent Yearn vault's `totalAssets` to return only assets managed by the vault
+   *   wrapper, rather than the parent Yearn vault.
+   */
+  function totalAssets() public view override(IEIP4626, BaseWrapper) returns (uint256) {
+    return _shareValue(totalSupply());
+  }
+
```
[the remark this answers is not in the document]
RedVeil: Thats a fair point. The original ERC4626 implements it exactly like we do. Not exactly sure what the intention was there but your suggestion seems better to me

anthonymartin (changes requested):

ecmendenhall:
@RedVeil @anthonymartin I fixed up the fuzz tests, fixed up the builds, and merged `develop` here. Check my work, but I think this beast is ready for a merge.

PR #524, opened 2022-03-07
Feature/sweet pool integration

RedVeil:
To test the frontend update the env variables as described in env.example and run `yarn hardhat node --network hardhat --tags zeroX-zapper` This creates all contracts necessary for the sweetVaults on the current block. One might need to update the addresses for hardhat in getNamedAccounts. If the node runs for too long zaps will fail since the blocknumbers are out of sync with the ZeroX API.


packages/app/components/NavBar/NavBar.tsx
```diff
@@ -20,13 +22,33 @@ export default function Navbar(): JSX.Element {
     }
   }, [chainId]);
 
+  const products: {
```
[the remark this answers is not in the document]
RedVeil: Could you than also add a specific `ProductLink` type or smth like that?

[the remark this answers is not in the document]
RedVeil: Also i just realized that with deeplinks we need to update the `url` and `currentlySelected` check to take networks into account. You can check this out in the TextLink.tsx file (``/${router?.query?.network}${url}`)

packages/app/components/NavBar/DropDownComponent.tsx
```diff
@@ -0,0 +1,48 @@
+import { Menu, Transition } from "@headlessui/react";
+import { Fragment } from "react";
+
+interface GetPopMenuProps {
+  options: Array<{ title: string; onClick: () => void; currentlySelected: boolean }>;
+}
+
+const DropDownComponent: React.FC<GetPopMenuProps> = ({ options }) => {
```
RedVeil, line 8: I like this generalized approach. Could we also use this for the network select?

packages/app/components/NavBar/DropDownComponent.tsx
```diff
@@ -0,0 +1,48 @@
...
+              {({ active }) => (
+                <div
+                  className={`group text-left md:flex md:flex-col md:justify-center md:text-center mb-4 md:mb-0 ml-2 md:ml-0 md:px-6 md:h-14 cursor-pointer md:border-b md:border-gray-200 last:border-0
+                  ${index === 0 ? "rounded-t-2xl" : ""} 
+                  ${length - 1 === index ? "rounded-b-2xl" : ""} 
+                  ${active ? "md:bg-gray-100" : "md:bg-white"} `}
+                  onClick={option.onClick}
+                >
+                  <p
+                    className={`whitespace-nowrap leading-none text-xl text-gray-500 md:text-gray-900 font-medium ${
```
RedVeil, line 32: The text seems a bit too big but i cant find any reference to the menu in the figma file. Could we change it to have the same size as the links in the navbar? Or point me to the figma file :)

packages/app/pages/sweetvaults.tsx
```diff
@@ -0,0 +1,113 @@
...
+import useWeb3 from "hooks/useWeb3";
+import React, { useMemo, useState } from "react";
+import ContentLoader from "react-content-loader";
+import toast, { Toaster } from "react-hot-toast";
+
+export default function index(): JSX.Element {
+  const { account, signer, onContractSuccess, onContractError, contractAddresses } = useWeb3();
+  const [searchValue, setSearchValue] = useState("");
+  const approveToken = useApproveERC20();
+  const { data: sweetVaults, mutate: revalidate } = useSweetVaults([
```
RedVeil, line 17: I think we should create an array of sweet vaults which gets returned in contractAddresses similar to staking. Than we can pass in the whole array without needing to update individual vaults here

packages/utils/src/getSweetVault.ts
```diff
@@ -0,0 +1,131 @@
...
+    symbol: vault.token.symbol,
+    icon: vault.icon,
+    pricePerShare: pricePerShare,
+    tvl: pricePerShare
+      .mul(totalSupply)
+      .div(parseEther("1"))
+      .mul(parseEther(vault.tvl.price.toString()))
+      .div(parseEther("1")),
+    apy: (vault?.apy?.net_apy / 2) * 100 * (98.5 / 100),
+    underlyingTokenContract: underlyingTokenContract,
```
RedVeil, line 121: I think we should call `getToken()` for each of the underlyingToken so we have all informations that we need about the underlying

RedVeil: @arne-z @Amirjab21 preferences?

packages/utils/src/getSweetVault.ts
```diff
@@ -0,0 +1,131 @@
...
+    deposited: BigNumber;
+    allowance: BigNumber;
+  };
+  underlyingTokenContract: ERC20;
+  contract: Vault;
+  curveLink: string;
+  infoText: InfoText;
+}
+
+const sweetVaults: { displayName: string; infoText: InfoText; link: string }[] = [
```
RedVeil, line 28: This should become its own type

packages/hardhat/deploy/10_SweetVaults.ts
```diff
@@ -0,0 +1,103 @@
...
+import { addContractToRegistry } from "./utils";
+
+async function giveDeployerUnderlyingToken(faucet, accounts, tokenAddress) {
+  switch (tokenAddress) {
+    case accounts.crvAave:
+      return await faucet.sendCrvAaveLPTokens(10, accounts.deployer);
+    case accounts.crvCvxCrv:
+      return await faucet.sendCrvCvxCrvLPTokens(10, accounts.deployer);
+    case accounts.dai:
+      return await faucet.send;
```
RedVeil, line 17: that looks unfinished

packages/hardhat/deploy/10_SweetVaults.ts
```diff
@@ -0,0 +1,103 @@
...
+  let faucet;
+  let contractRegistry;
+  if (["hardhat", "local"].includes(hre.network.name)) {
+    faucet = await hre.ethers.getContractAt("Faucet", (await deployments.get("Faucet")).address, signer);
+  }
+
+  const vaultStakingPools = await getVaultStakingPools(hre.network.config.chainId, addresses, deployments);
+
+  const FEE_MULTIPLIER = parseEther("0.0001"); // 1e14
+  const feeStructure = {
```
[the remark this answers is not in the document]
RedVeil: Id also rename `feeStructure` into `FEE_STRUCTURE` to make clear that it shouldnt be changed

packages/hardhat/deploy/10_SweetVaults.ts
```diff
@@ -0,0 +1,103 @@
...
+      from: addresses.deployer,
+      args: [vaultStakingPools[i].rewardsToken, VaultDeployed.address, addresses.rewardsEscrow],
+      log: true,
+      autoMine: true,
+      contract: "Staking",
+    });
+
+    const vault = await ethers.getContractAt("Vault", VaultDeployed.address);
+
+    await deploy("VaultFeeController", {
```
RedVeil, line 65: We only need to create one VaultFeeController so that can be take out of the loop. The VaultFeeController is one central contract that controls fees for all vaults.

packages/hardhat/deploy/10_SweetVaults.ts
```diff
@@ -0,0 +1,103 @@
...
+          management: FEE_MULTIPLIER.mul(200),
+          performance: FEE_MULTIPLIER.mul(2000),
+        },
+        addresses.contractRegistry,
+      ],
+      log: true,
+      autoMine: true,
+      contract: "VaultFeeController",
+    });
+    await addContractToRegistry("VaultFeeController", deployments, signer, hre);
```
RedVeil, line 82: This should therefore also just be called once.

packages/app/pages/sweetvaults.tsx
```diff
@@ -0,0 +1,113 @@
...
+      `${sweetVault?.name} approved!`,
+      () => revalidate(),
+    );
+  }
+
+  return (
+    <div className="w-full h-full">
+      <Navbar />
+      <Toaster position="top-right" />
+      <div className="lg:w-11/12 lglaptop:w-9/12 2xl:max-w-7xl lg:mx-auto mx-6 md:mx-8 mt-10 lg:mt-14 pb-6">
```
RedVeil, line 66: This is not aligned with the navbar and sticks for me to the left side instead of beeing centered.

packages/hardhat/lib/utils/getStakingPools.ts
```diff
@@ -69,6 +130,18 @@ export async function getStakingPools(chainId: number, addresses, deployments):
           inputToken: butter,
           rewardsToken: (await deployments.get("TestPOP")).address,
         },
+        {
```
RedVeil, line 133: Same here

packages/hardhat/lib/utils/getContractAddresses.ts
```diff
@@ -19,6 +19,8 @@ const butterDependencyContractNames = [
 
 const stakingContractNames = ["butterStaking", "popUsdcLpStaking"];
```
RedVeil, line 20: We should add the names of the sweetVault staking contracts

packages/app/pages/sweetvaults.tsx
RedVeil: My bad! You already did that! :) So lets just use`contractAddresses.sweetVaults` for the hook

packages/app/pages/sweetvaults.tsx
```diff
@@ -0,0 +1,113 @@
...
+    contractAddresses.daiSweetVault,
+    contractAddresses.threeCrvSweetVault,
+  ]);
+
+  const filteredArray = useMemo(() => {
+    return sweetVaults?.filter((option) => option.name.toLowerCase().includes(searchValue.toLowerCase()));
+  }, [searchValue, sweetVaults]);
+
+  function deposit(amount: BigNumber, sweetVault: SweetVaultMetadata): void {
+    toast.loading(`Depositing ${sweetVault?.name} ...`);
```
RedVeil, line 27: I think we should rename this to `Depositing ${underlyingToken.name}`

packages/app/pages/sweetvaults.tsx
```diff
@@ -0,0 +1,113 @@
...
+      .then((res) =>
+        onContractSuccess(res, `Deposited ${sweetVault?.name}!`, () => {
+          revalidate();
+        }),
+      )
+      .catch((err) => onContractError(err));
+  }
+
+  function withdraw(amount: BigNumber, sweetVault: SweetVaultMetadata): void {
+    toast.loading(`Withdrawing p${sweetVault?.name} ...`);
```
RedVeil, line 40: Same here

packages/app/pages/sweetvaults.tsx
```diff
@@ -0,0 +1,113 @@
...
+      .then((res) =>
+        onContractSuccess(res, `${sweetVault?.name} withdrawn!`, () => {
+          revalidate();
+        }),
+      )
+      .catch((err) => onContractError(err));
+  }
+
+  function approve(sweetVault: SweetVaultMetadata): void {
+    toast.loading(`Approving ${sweetVault?.name} ...`);
```
RedVeil, line 53: And here :D

packages/app/components/NavBar/DropDownComponent.tsx
Amirjab21: Would it improve it?

currently network select is this:

`interface NetworkOptionsMenuProps {
  currentChain: number;
  switchNetwork: (chainId: number) => void;
}`

packages/app/components/NavBar/DropDownComponent.tsx
Amirjab21: Changed to text-lg

Screenshot 2022-05-06 at 14 15 53

packages/hardhat/deploy/10_SweetVaults.ts
Amirjab21: we already give the deployer DAI and 3CRV, I can complete it anyway

packages/utils/src/getSweetVault.ts
Amirjab21: @RedVeil not sure what you mean by this - don't we already have all the details about underlying token when we connect to the contract?

packages/app/components/SweetVaults/SweetVaultsDepositInterface.tsx
```diff
@@ -79,8 +96,10 @@ const SweetVaultsDepositInterface: React.FC<SweetVaultProps> = ({ sweetVault, de
           <div className="mt-10">
             <MainActionButton
               label="Withdraw"
-              handleClick={() => withdraw(amount, sweetVault)}
-              disabled={amount.eq(BigNumber.from("0")) || amount.gt(sweetVault.balances.deposited)}
+              handleClick={() => {
+                withdraw(amount, sweetVault);
+              }}
+              disabled={amount.eq(BigNumber.from("0")) || amount.gt(sweetVault.vault.balance)}
```
Amirjab21, line 102: This looks like a mistake to me?

sweetVault.vault.balance = underlyingTokenContract.balanceOf(userAddress);

This is the users balance. When withdrawing, shouldn't the requirement be that the vault has greater than the balance requested by the user?

RedVeil: vault essentially is now of type `Token`

packages/hardhat/lib/utils/getStakingPools.ts
```diff
@@ -96,18 +95,6 @@ export async function getStakingPools(chainId: number, addresses, deployments):
           inputToken: butter,
           rewardsToken: (await deployments.get("TestPOP")).address,
         },
-        {
```
Amirjab21: To confirm - the only way to access these vault staking contracts is through the actual getVaultStakingPools function which can return the vaultName and the stakingPoolName

RedVeil: thats what i was thinking. otherwise we deploy the staking contract for vaults twice. once in the staking deploy script and once in the vaults deploy script

PR #552, opened 2022-03-18
Feature/deposit and stake

yaoandrew:
- Adds stub and interface for `depositAndStake` and `mintAndStake`


packages/hardhat/contracts/core/defi/vault/Vault.sol
```diff
@@ -262,6 +266,38 @@ contract Vault is IEIP4626, AffiliateToken, ReentrancyGuard, Pausable, ACLAuth,
...
+    whenNotPaused
+    blockLocked
+    onlyApprovedContractOrEOA
+    returns (uint256)
+  {
+    return _depositAndStakeFor(assets, msg.sender);
+  }
+
+  /**
+   * @notice Deposit exactly `assets` amount of tokens, issuing vault shares to receiver and staking in the staking contract.
```
ecmendenhall, line 286:
```suggestion
   * @notice Deposit exactly `assets` amount of tokens, issuing vault shares to `receiver` and staking in the staking contract.
```

PR #554, opened 2022-03-21
Feature/btrv2 integration

RedVeil:
Tested Integration of BTR-V2. Frontend works as expected. Added minor display improvements.

 -- One caviat of the hooks is that the butterBatchData reloads everytime an on chain interaction has happened. Everytime this happens this Mint/Redeem Interface will switch back to the default view of minting BTR with 3CRV. This means that if a user approves BTR or USDC for example they will need to go back to their tab/ currency from the default view. But since we need this useEffect to display everything correctly on inital load im not sure how to change this.
  -> I would keep this in mind and currently roll with it. If we find a better way in the future we should certainly change this.
  
  -- When minting BTR their is only on the first time a modal guiding you what to do. On redeem their is a modal that pops up everytime. Since we dont really have a way of knowing if a user minted BTR before i think we should just show the modal for redemption also only once.


RedVeil:
Ticket https://github.com/popcorndao/sweet-caramel/pull/554

RedVeil:
to my comments above:
- I added an initalLoad boolean which gets set when data gets loaded the first time. Only on initalLoad we set the input token to 3crv and ouput to butter in all other cases we simply utilize the MintRedeemToggle

- For the modals i changed the system to now have one mint and one redeem modal which always show if the user doesnt click "dont remind me again"

RedVeil:
- Est. APY shows now the yield from BTR + APY from butter staking.
- Added a Tooltip and extra line of text explaining how to get the full apy.
- Mint and redeem modal are now essentially the same and point the user to the table below to check their batches. They  can both be closed once or the user can choose to not be reminded again.
- SelectedToken uses keys now. Therefore we can update the data on each token while still remembering which token a user had selected after an onchain action.

arne-z (approved):

packages/app/pages/butter.tsx
```diff
@@ -593,8 +650,8 @@ export default function Butter(): JSX.Element {
                                 </h3>
                                 <div className="my-4">
                                   <p className="text-lg text-gray-500">
-                                    The Butter redemption process involves unwrapping Butter's underlying assets into
-                                    stablecoins.
+                                    Redeem your BTR to receive its value in 3CRV or stablecoins. We will convert all the
+                                    underlying token back into 3CRV or your desired stablecoin.
```
anthonymartin, line 654:
```suggestion
                                    underlying tokens back into 3CRV or your desired stablecoin.
```

packages/app/pages/butter.tsx
```diff
@@ -580,10 +637,10 @@ export default function Butter(): JSX.Element {
                                 </h3>
                                 <div className="my-4">
                                   <p className="text-lg text-gray-500">
-                                    Butter is a token that represents a yield accrual strategy. The minting process
-                                    involves converting deposited stablecoins into other stable assets that are
-                                    compatible with the yield accrual strategy. As the value of the underlying assets
-                                    increase, so does the redeemable value of Butter.
+                                    Mint BTR with 3CRV or stablecoins to earn interest on multiple Stablecoins at once.
```
anthonymartin, line 640: i like this copy change here

packages/app/components/BatchButter/MintRedeemToggle.tsx
```diff
@@ -28,7 +28,7 @@ const MintRedeemToggle: React.FC<MintRedeemToggleProps> = ({ redeeming, setRedee
               classExtras="h-7 w-7 mt-0 ml-5"
               id="1"
               title="Mint"
-              content="Butter is a token that represents a yield accrual strategy. The minting process involves converting deposited stablecoins into other stable assets that are compatible with the yield accrual strategy. As the value of the underlying assets increase, so does the redeemable value of Butter."
+              content="Mint BTR with 3CRV or stablecoins to earn interest on multiple Stablecoins at once."
```
anthonymartin, line 31: Do you want to pull in the copy from below that I commented on here so this is more explanatory?

anthonymartin (changes requested):

packages/app/components/BatchButter/MintRedeemToggle.tsx
RedVeil: sry im a little confused which copy from where? Do you mean adding this back in: "The minting process involves converting deposited stablecoins into other stable assets that are compatible with the yield accrual strategy."

packages/app/components/BatchButter/MintRedeemToggle.tsx
```diff
@@ -28,7 +28,10 @@ const MintRedeemToggle: React.FC<MintRedeemToggleProps> = ({ redeeming, setRedee
               classExtras="h-7 w-7 mt-0 ml-5"
               id="1"
               title="Mint"
-              content="Butter is a token that represents a yield accrual strategy. The minting process involves converting deposited stablecoins into other stable assets that are compatible with the yield accrual strategy. As the value of the underlying assets increase, so does the redeemable value of Butter."
+              content="Mint BTR with 3CRV or stablecoins to earn interest on multiple Stablecoins at once.
+              As the value of the underlying assets increase, so does the redeemable value of
+              Butter. This process converts deposited funds into other stablecoins and deploys
+              them in Yearn to generate interest."
```
anthonymartin, line 34:
```suggestion
              them to automated yield-farming contracts by Yearn to generate interest."
```

anthonymartin (approved):

PR #560, opened 2022-03-22
locked staking details added for POP

Amirjab21:
Main things to check here for the devs -->

1) Whether the decision to create a new state variable for the selected vesting record is the right one - I decided to make a new state piece as its a bit disconnected to the other pieces of data held in the form state.

2) Whether the useMemo should be used on popLockerInteraction.tsx or if this manipulation should be done on the file that gets the staking pool info.


packages/app/components/staking/PopLockerInteraction.tsx
```diff
@@ -24,16 +27,61 @@ export default function PopLockerInteraction({
   const withdrawal = type === InteractionType.Withdraw;
   const deposit = type === InteractionType.Deposit;
   const stakingToken = stakingPool?.stakingToken;
+  const lockedBalances = useMemo(
```
RedVeil, line 30: I think i would move this to https://github.com/popcorndao/sweet-caramel/blob/3b5f0099ac5040ce63fc499e47574b9c62b91896/packages/utils/src/getStakingPool.ts#L45. This means we could also get rid of the memo. All necessary data for display come than again from one source which is `getPopLocker()`. It would entail do add an optional type on StakingPool of lockedBalances. Its a great question though. Had to think about it a bit since this here is also quite clean. In the end i prefer to have all data coming from one place so we can change it in one place and dont have to search for multiple files.

RedVeil (approved):

PR #565, opened 2022-03-25
Feature/support deeplinks include chain in uri and modal to switch chain when wallet doesn't match url.

arne-z:


arne-z:
Superceded by https://github.com/popcorndao/sweet-caramel/pull/608

PR #595, opened 2022-04-13
permit functionality

Amirjab21:


packages/app/hooks/tokens/useERC20Permit.ts
```diff
@@ -0,0 +1,23 @@
+import { ERC20Permit__factory } from "@popcorn/hardhat/typechain";
+import { getToken } from "@popcorn/utils";
+import { useEffect, useState } from "react";
+import { Token } from "../../../utils/src/types/index";
+import useWeb3 from "../useWeb3";
+
+export default function useERC20Permit(address: string | null, permit: boolean = false): Token {
```
RedVeil, line 7: I think we should generalize this further and just have an `useERC20`. The `useERC20` should have a try/catch in which it instantiates the address as an ERC20_permit and calls `DOMAIN_SEPARATOR`. I checked an solmate, openZeppelin, usdc, dai, yearn all of them use it to construct their permit. If this try/catch call fails we should just return a normal ERC20.

packages/app/hooks/tokens/useERC20Permit.ts
```diff
@@ -0,0 +1,23 @@
...
+  const { signerOrProvider, account } = useWeb3();
+  const [token, setToken] = useState<Token>(null);
+  useEffect(() => {
+    let mounted = true;
+    if (address && signerOrProvider) {
+      getToken(ERC20Permit__factory.connect(address, signerOrProvider))
```
RedVeil, line 13: We should also pull in getToken from `sweet-pool-integration`. In that getToken we add allowance to the object.
This would need to be upgraded though so that the getToken call uses the `useAllowance`-hook. The `useAllowance`-hook should than take care of allowance or/and permit depending on the type of input token.

RedVeil: Oh also we need to create our own interface i think. It should have two/three permit functions simply so we can cover the different permit implementations via one interface and factory.

packages/app/pages/[network]/butter/index.tsx
```diff
@@ -527,13 +548,52 @@ export default function Butter(): JSX.Element {
...
+    if (
+      getCurrentContractAddress() === butterBatchZapper.address &&
+      (contractKey === "dai" || contractKey === "usdc")
+    ) {
+      const valueAdjusted =
+        butterPageState.selectedToken.input === "usdc"
+          ? butterPageState.depositAmount.div(1e12)
+          : butterPageState.depositAmount;
+
+      await getSignature(
```
Comment (RedVeil) [inline packages/app/pages/[network]/butter/index.tsx:560]: With this new change all permits/permissions should be done via `useApproveERC20`. This means that `useApproveERC20` needs to be updated to call permit in case the function exists on the token. (We might need some mapping here or smth else to make sure we call the right kind of permit for the right tokens. Open for suggestions how to do this in a clean way)

packages/app/hooks/tokens/useERC20Permit.ts
Amirjab21: How would the useAllowance hook get the allowance of a permit call? This only exists as an off-chain signature that we store on the front-end

PR #605, opened 2022-04-20
feature/ adjust mobile breakpoint from 768 to 1024

Ajiva-D:


arne-z (changes requested):
Code wise it all looks good but there are a few unintended consequences happening that you can find when looking through the deploy preview.
image
The Disconnect button doesn't like being stretched this wide, I think it maybe should keep the icon centered next to the text.
image
The happy staking image needs to either grow in length to accommodate the text, or the image can't grow quite this large.
A similar thing as with the disconnect button happens with the icons here. I'd also suggest adding a bit more bottom padding to the mobile menu, it's a slight bit weird having the menu stuck to the bottom of the screen unable to scroll further past it.
image

arne-z (approved):
👌

PR #609, opened 2022-04-22
updated hardhat, removed forked hardhat-deploy for main hardhat-deploy

RedVeil:


arne-z (approved):
Solves the issue for me.

PR #628, opened 2022-04-26
Feature/beneficiaries footer

Ajiva-D:


RedVeil:
The footer breaks weirdly at around 1010px. I think this could be solved by using the same screen settings in tailwind as in /app. Than it would be md and therefore would use the mobile footer

RedVeil:
Could you set `target:"_blank"` for all links in the "Connect" column and for the for links that use icons (fb,twitter,github and discord). Since these are external sites i think we should open these pages in a new tab

Ajiva-D:
Resolved both comments

PR #632, opened 2022-04-27
Feature/beneficairies top nav

Ajiva-D:


packages/grants/components/Header.tsx
```diff
@@ -0,0 +1,135 @@
...
+            <ChevronDownIcon className="w-5 h-5" aria-hidden="true" />
+            <BeneficiaryFilter
+              filterList={filterList}
+              switchFilter={switchFilter}
+              position="absolute top-10 left-1/2 transform -translate-x-1/2"
+              width="w-60"
+            />
+          </Menu.Button>
+        </Menu>
+        <p className="text-gray-500 font-semibold hidden lg:block">Eligible Beneficiaries</p>
```
RedVeil, line 48: we should make this into a link already for later

packages/grants/components/Header.tsx
```diff
@@ -0,0 +1,135 @@
...
+        <p className="text-gray-500 font-semibold hidden lg:block">FAQ</p>
+      </div>
+
+      <div className="pr-5 md:pr-10 flex items-center gap-5 md:gap-10">
+        <Button variant="primary" className="md:w-44 py-3 px-8 md:px-0">
```
RedVeil, line 53: I think we should create some general button component to always use the same design here without much code duplication. Additionally ill ask Fan what kind of hover styling these buttons should have.

RedVeil:
Love how the buttons change when going to tablet and mobile

RedVeil:
The mobile menu on tablet and mobile needs some px/mx to be aligned with the popcorn icon

packages/grants/components/Header.tsx
Ajiva-D: This button is a component with different variants. From the whole design I noticed that the button widths and paddings varied, that's why I added className as a prop to style those things differently. But other general styles are in the Button Component.

packages/grants/components/Header.tsx
Ajiva-D: Done

PR #663, opened 2022-05-02
Feature/beneficiary application step 3

Ajiva-D:


Ajiva-D:
I removed the "upload files" button that is the design from the implementation cos it's not needed. When the file is selected from the input, it starts uploading to IPFS, we don't need another button to upload.

packages/grants/components/CommonComponents/IPFSUploadFunc.tsx
```diff
@@ -0,0 +1,156 @@
...
+import React, { useEffect, useState } from "react";
+import { useDropzone } from "react-dropzone";
+import toast from "react-hot-toast";
+import * as SVGLoaders from "svg-loaders-react";
+
+const success = (msg: string) => toast.success(msg);
+
+const uploadError = (errMsg: string) => toast.error(errMsg);
+const isSuccessfulUpload = (res: UploadResult): boolean => res.status === 200;
+const isFailedUpload = (res: UploadResult): boolean => res.status !== 200;
```
arne-z, line 11: It seems that UploadResult doesn't expose a .ok entry, otherwise I'd recommend using that here. Should we be concerned about 2xx results other than 200 occurring and falsely getting flagged as errors?

packages/grants/components/CommonComponents/IPFSUploadFunc.tsx
```diff
@@ -0,0 +1,156 @@
...
+      message: `File is larger than ${maxFileSizeMB} MB`,
+    };
+  }
+  return null;
+};
+
+const showUploadBox = (numMaxFiles: number, localState: string | string[]): boolean => {
+  if (typeof localState === "string") return localState === "";
+  return localState.length < numMaxFiles || numMaxFiles === 0;
+};
```
arne-z, line 36: I'm not sure I fully understand this check.
The upload box is displayed if the local state is of string type and empty or if it is of array type and there is still "space" for more files to be uploaded right?
The second half of this makes sense but I'm not sure why the first check is the way it is.
This might be a good place to leave a comment explaining this.

packages/grants/components/CommonComponents/SocialMedia.tsx
```diff
@@ -52,7 +52,7 @@ const SocialMedia: React.FC<SocialMediaProps> = ({ beneficiary }) => {
     <>
       <div className="flex space-x-6">
         <SocialMediaIcon url={beneficiary?.links?.website} platform={"website"} />
-        <SocialMediaIcon url={beneficiary?.links.contactEmail} platform={"email"} />
+        <SocialMediaIcon url={beneficiary?.links.contactEmail.data} platform={"email"} />
```
arne-z: Are we certain beneficiary?.links.contactEmail will always be defined? This will throw a typeerror if it's trying to access .contactEmail of undefined, so it might be good to add more "?" here.
```suggestion
        <SocialMediaIcon url={beneficiary?.links?.contactEmail?.data} platform={"email"} />
```

packages/grants/components/Proposals/Form/Intro.tsx
```diff
@@ -159,7 +159,7 @@ const Intro: React.FC<FormStepProps> = ({ form, navigation, visible }) => {
               </h2>
               <div className="mt-6 space-y-4 sm:space-y-0 sm:flex sm:space-x-5">
                 {/* Check for partially completed form */}
-                {formData.organizationName !== "" && (
+                {formData.organizationName.data !== "" && (
```
arne-z: I wonder if this should either be an explicit check for data to be defined and empty or a check for data to be truthy, as right now this will be true on undefined or defined and having content.

packages/grants/styles/globals.css
```diff
@@ -117,3 +117,11 @@ p {
...
 	width:1px
 }
+.apply-card{
+	padding-top: 400px;
+}
+@media screen and (max-width:767px){
+	.apply-card{
+		padding-top: 800px;
+	}
+}
```
arne-z, line 127:
```suggestion
}

```

arne-z (changes requested):

packages/grants/components/Apply/CoverPhotoUpload.tsx
```diff
@@ -0,0 +1,45 @@
...
+  return (
+    <IPFSUploadFunc {...props}>
+      <CoverImg
+        className=" bg-gray-100 rounded-2xl flex flex-col justify-center items-center py-16 relative h-44"
+        bgImage={`https:/popcorn.mypinata.cloud/ipfs/${props.localState}`}
+      >
+        {!props.localState && (
+          <>
+            <p className=" font-semibold text-blue-600">
+              {" "}
```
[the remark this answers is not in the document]
Ajiva-D: definitely formatting

packages/grants/components/Apply/FlowSteps.tsx
```diff
@@ -0,0 +1,41 @@
+import React from "react";
+interface FlowStepsProps {
+  active: Number;
```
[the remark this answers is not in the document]
Ajiva-D: Steps are enum, this is a prop to pass the current step. The enum is declared in /pages//apply/form

packages/grants/components/Apply/GeneralInformation.tsx
```diff
@@ -0,0 +1,308 @@
...
+          image,
+          description: formData?.files?.headerImage?.description,
+        },
+      },
+    });
+  };
+
+  return (
+    <div>
+      {" "}
```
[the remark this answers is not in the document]
Ajiva-D: formatting

arne-z (approved):

PR #711, opened 2022-05-10
0x Vault Zapper

ecmendenhall:
A POC vault zapper using the Zapper Curve zap contracts. See the annotated test case for all the details.


packages/hardhat/lib/tasks/curve/metapoolInterfaces.ts
```diff
@@ -0,0 +1,104 @@
...
+      } else {
+        console.log("JSON saved");
+      }
+    });
+  }
+
+  return pools;
+}
+
+export default task("curve:metapool-interfaces", "get all curve metapool interfaces and save them to a file")
```
anthonymartin, line 100: nice

packages/hardhat/lib/tasks/forge/forgeTest.ts
```diff
@@ -0,0 +1,28 @@
...
+  try {
+    const { stdout, stderr } = await execCommand(command);
+    console.log(stdout);
+    console.log("stderr:", stderr);
+  } catch (e) {
+    console.error(e); // should contain code (exit code) and signal (that caused the termination).
+  }
+}
+
+export default task("forge:forge-test", "run forge tests")
```
anthonymartin, line 21: this is cool. we can also use the `scripts` property in `package.json` to call scripts dynamically as well without having to write as much code

anthonymartin (approved):

packages/hardhat/contracts/core/defi/ZeroXZapper.sol
```diff
@@ -0,0 +1,270 @@
...
+      intermediateToken,
+      toToken,
+      minToTokens,
+      swapTarget,
+      swapCallData,
+      address(this),
+      false
+    );
+
+    IERC20(toToken).safeTransfer(_getContract(FEE_CONTROLLER_ID), amountOut);
```
anthonymartin, line 166: https://github.com/popcorndao/sweet-caramel/blob/feature/vault-zapper-zerox/packages/hardhat/contracts/core/defi/vault/VaultFeeController.sol

There's nothing in this contract to transfer fees out. I think we should replace this with a treasury that's defined in this contract itself

anthonymartin (changes requested):

packages/hardhat/lib/adapters/Zapper.ts
```diff
@@ -0,0 +1,221 @@
...
+
+  public async zapOut(
+    to: Token,
+    vault: Vault,
+    stableSwapAddress: string,
+    sellAmount: BigNumber,
+    slippagePercentage: number
+  ) {
+    const curveLPAddress = await vault.asset();
+    const [coin0, coin1] = await this.getCoins(curveLPAddress, vault.provider);
```
anthonymartin, line 55: https://curve.readthedocs.io/registry-registry.html#getting-info-about-a-pool

we could use the registry to `get_coins` which should work for all pools

ecmendenhall:
Fixed up the Forge test fork block here. 

The Forge tests are now wrapped in a `hardhat forge:forge-test` task that lets us specify different blocks per test file [here](https://github.com/popcorndao/sweet-caramel/blob/90bc8dd407d610201db8548c23431e942f759477/packages/hardhat/lib/tasks/forge/forgeTest.ts#L9). For now, there is a simple filename/block mapping in the task file, but we can extract this config somewhere else if we want as it grows. There are similar wrapper scripts defined in `package.json`: `yarn test:forge` in this package will invoke the Hardhat task and `yarn test:forge` at the project root will run it with Lerna.

This was the best out of a bunch of bad options I explored, but I think it will work for now. Switching fork blocks is coming Any Day Now to Foundry, but it's not quite ready.

packages/hardhat/contracts/core/defi/ZeroXZapper.sol
```diff
@@ -0,0 +1,270 @@
...
+    address intermediateToken,
+    address toToken,
+    uint256 minToTokens,
+    address swapTarget,
+    bytes calldata swapCallData
+  ) public onlyRole(DAO_ROLE) {
+    uint256 feeBal = fees[vaultAsset].accumulated;
+    fees[vaultAsset].accumulated = 0;
+
+    IERC20(vaultAsset).approve(address(zapperOut), feeBal);
```
ecmendenhall, line 152: We know here that `vaultAsset` is always a Curve LP token, right? Thinking about whether we need to zero approvals first or use `safeApprove`/`increaseAllowance`.

packages/hardhat/contracts/core/defi/ZeroXZapper.sol
```diff
@@ -0,0 +1,270 @@
...
+
+      uint256 balanceBefore = fromToken.balanceOf(address(this));
+      fromToken.safeTransferFrom(msg.sender, address(this), incomingTokenQty);
+      uint256 balanceAfter = fromToken.balanceOf(address(this));
+
+      amountReceived = balanceAfter - balanceBefore;
+      fromToken.safeApprove(address(zapperIn), 0);
+      fromToken.safeApprove(address(zapperIn), amountReceived);
+    }
+    uint256 amountOut = zapperIn.ZapIn{ value: msg.value }(
```
ecmendenhall, line 196: It's possible for the caller to provide native ETH accidentally even when they intend to do an ERC20 transfer. We should probably `require(msg.value == 0)` in the `if` statement above so that providing ETH + an ERC20 address reverts.

packages/hardhat/contracts/core/defi/ZeroXZapper.sol
```diff
@@ -0,0 +1,270 @@
...
+    address swapTarget,
+    bytes calldata swapCallData
+  ) public {
+    address vault = vaults[vaultAsset];
+    require(vault != address(0), "Invalid vault");
+
+    IVault(vault).redeem(amount, address(this), msg.sender);
+
+    // For some reason the value returned of redeem is sometimes a few WEI off which is why i opted for this solution
+    uint256 withdrawn = IERC20(vaultAsset).balanceOf(address(this));
```
ecmendenhall, line 231: Should we use the same `balanceBefore`/`balanceAfter` pattern here?

packages/hardhat/contracts/core/defi/ZeroXZapper.sol
anthonymartin: even better!

packages/hardhat/contracts/core/defi/ZeroXZapper.sol
```diff
@@ -0,0 +1,270 @@
...
+  }
+
+  /**
+   * @notice Changes the fee for a certain vault asset
+   * @param vaultAsset Addres of the underlying asset of a vault
+   * @param inBps DepositFee in BPS
+   * @param outBps WithdrawlFee in BPS
+   * @dev Per default both of these values are not set. Therefore a fee has to be explicitly be set with this function
+   */
+  function setFee(
```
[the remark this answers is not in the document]
RedVeil: Done

packages/hardhat/contracts/core/defi/ZeroXZapper.sol
```diff
@@ -0,0 +1,270 @@
...
+
+  struct Fee {
+    uint256 accumulated;
+    uint256 inBps;
+    uint256 outBps;
+  }
+
+  /* ========== STATE VARIABLES ========== */
+
+  IZapIn constant zapperIn = IZapIn(0x5Ce9b49B7A1bE9f2c3DC2B2A5BaCEA56fa21FBeE);
```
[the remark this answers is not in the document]
RedVeil: I liked this to make it a little more explicit

packages/hardhat/contracts/core/defi/ZeroXZapper.sol
RedVeil: i think we are good with that. In this zapper they have to be curve LP token. Otherwise we wouldnt be able to use these ZapIn and ZapOuts.

packages/hardhat/contracts/core/defi/ZeroXZapper.sol
```diff
@@ -112,29 +120,54 @@ contract ZeroXZapper is ACLAuth, ContractRegistryAccess {
...
 
-  function addVault(address underlyingToken, address vault) external {
+  function updateVault(address underlyingToken, address vault) external {
     _requireRole(DAO_ROLE);
     vaults[underlyingToken] = vault;
+    emit UpdatedVault(underlyingToken, vault);
+  }
+
+  function removeVault(address underlyingToken) external {
+    _requireRole(DAO_ROLE);
```
ecmendenhall, line 130:
```suggestion
  function removeVault(address underlyingToken) external onlyRole(DAO_ROLE) {
```

anthonymartin (approved):

PR #723, opened 2022-05-15
Feature/keeper incentives updates

anthonymartin:
The main reasons for this update are to have separate budgets/balances for incentives (before only global budget/balance for keeper incentives was maintained), and to enforce a cooldown period for all keeper incentives so that balances can't be drained by calling a keeper incentivized function consecutively. Additionally, several views are added to easily view Incentive account balances and keeper account balances, since incentives support multiple payment tokens and are now claimed instead of transferred on each job invocation.

* adds claimable keeper incentives instead of transfering tokens for
  each successful job
* supports multiple incentive tokens
* moves burn logic to incentive struct
* adds view to return claimable balances
* adds view to return keeperIncentive configurations
* claiming from a incentiveAccount will clear out the account balance
* adds claimFor for claiming
* budget is set on per incentive basis instead of globally
* incentives are now owned by addresses and not contract names
* adds ButterBatchProcessingV1, IncentivedV1 and KeeperIncentiveV1 for fork test backwards compatability
* adds helpers for keeper incentives
* incentives each have their own cooldown period

## Checklist before requesting a review

- [x] I have crossed off all todos in the issue ticket.
- [x] There are no more TODOS in the PR.
- [x] There are no more console.logs for debugging or other temporary functions and comments.
- [x] Functions and variables have clear, distinct names.
- [x] I performed a self review of the code.
- [x] All tests are passing.
- [x] The build and ci checks are successful.

For Smart Contracts:
- [x] We have 100% code coverage.
- [x] Deploy scripts are included and updated.
- [x] Fuzz tests are implemented when applicable.
- [x] Forge tests are added/updated.
- [x] All functions have natspec comments.


packages/hardhat/contracts/core/utils/KeeperIncentive.sol
```diff
@@ -16,32 +16,96 @@ contract KeeperIncentive is ACLAuth, ContractRegistryAccess {
...
+  mapping(address => uint256) public keeperClaims;
+
+  /**
+   * pending burn balance
+   */
+  uint256 public burnBalance;
+
+  uint256 public incentivesAccountsCount;
   address internal immutable burnAddress = 0x000000000000000000000000000000000000dEaD; // Burn Address
   uint256 public requiredKeeperStake;
```
bhanusanghi, line 84: Should KeeperStake be global or at Incentive level ? Since rewards vary with incentive, should stake vary too with incentive.

packages/hardhat/contracts/core/utils/KeeperIncentive.sol
```diff
@@ -58,97 +122,365 @@ contract KeeperIncentive is ACLAuth, ContractRegistryAccess {
...
+    bytes32 incentiveAccountId = _incentiveAccountId(_contractAddress, _i, _rewardToken);
     incentive.reward = _reward;
     incentive.enabled = _enabled;
     incentive.openToEveryone = _openToEveryone;
-    emit IncentiveChanged(_contractName, oldReward, _reward, oldOpenToEveryone, _openToEveryone);
+    incentive.rewardToken = _rewardToken;
+    incentive.cooldown = _cooldown;
+    incentive.burnRate = _burnRate;
+    incentiveAccountRewardsToken[incentiveAccountId] = _rewardToken;
+    rewardsTokensIncentiveAccounts[_rewardToken].push(incentiveAccountId);
```
bhanusanghi, line 403: When incentive is updated what happens to existing incentive Balance? It should be redirected to new incentiveAccountId.

packages/hardhat/contracts/core/utils/KeeperIncentive.sol
```diff
@@ -16,32 +16,96 @@ contract KeeperIncentive is ACLAuth, ContractRegistryAccess {
     uint256 reward; //pop reward for calling the function
     bool enabled;
     bool openToEveryone; //can everyone call the function to get the reward or only approved?
+    address rewardToken;
+    uint256 cooldown; // seconds required between incentive payouts
+    uint256 burnRate;
```
bhanusanghi, line 21: As it is not a function of time, should it be renamed to burnPercentage ?

bhanusanghi (changes requested):
One Last comment - Should controllers have access to only incentives they are associated with? Currently no such check. We only check if the call handleKeeperIncentive is coming from a controller contract. 
This limits attack possibilities to only incentiveFunds associated with that controllerContract.

packages/hardhat/contracts/core/utils/KeeperIncentive.sol
anthonymartin: true. good point

packages/hardhat/contracts/core/utils/KeeperIncentive.sol
```diff
@@ -58,97 +122,365 @@ contract KeeperIncentive is ACLAuth, ContractRegistryAccess {
...
+    emit Burned(burnAmount);
   }
 
-  function fundIncentive(uint256 _amount) external {
-    IERC20(_getContract(keccak256("POP"))).safeTransferFrom(msg.sender, address(this), _amount);
-    incentiveBudget = incentiveBudget + _amount;
-    emit IncentiveFunded(_amount);
+  function _burn(uint256 _amount, bytes32 incentiveAccountId) internal {
+    incentiveBalances[incentiveAccountId] -= _amount;
+    burnBalance = _amount;
```
ecmendenhall, line 445: Should this increment `burnBalance` rather than setting it?

packages/hardhat/contracts/core/utils/KeeperIncentive.sol
```diff
@@ -160,11 +492,6 @@ contract KeeperIncentive is ACLAuth, ContractRegistryAccess {
     burnRate = _burnRate;
```
ecmendenhall, line 492: Validate here too.

packages/hardhat/contracts/core/utils/KeeperIncentive.sol
```diff
@@ -16,32 +16,96 @@ contract KeeperIncentive is ACLAuth, ContractRegistryAccess {
...
+
+  /**
+   * @notice mapping of account to rewardToken => balance
+   */
+  mapping(address => mapping(bytes32 => uint256)) public keeperBalances;
+
+  /**
+   * @notice bytes32 key is incentive id: keccak256(abi.encode(_contractName, _i))
+   */
+  mapping(bytes32 => uint256) public lastPayouts;
```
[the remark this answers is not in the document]
K-G-J: now using lastInvocation in Incentive struct

packages/hardhat/contracts/core/utils/KeeperIncentive.sol
```diff
@@ -58,97 +122,365 @@ contract KeeperIncentive is ACLAuth, ContractRegistryAccess {
...
+
+  function _payoutIncentive(
+    address _keeper,
+    Incentive memory incentive,
+    bytes32 incentiveAccountId
+  ) internal {
+    lastPayouts[incentiveAccountId] = block.timestamp;
+    (uint256 payoutAmount, uint256 burnAmount) = _previewPayout(incentive);
+    _deposit(_keeper, payoutAmount, incentiveAccountId);
+    if (incentive.burnRate > 0) {
```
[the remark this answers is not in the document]
K-G-J: done

packages/hardhat/contracts/core/utils/KeeperIncentive.sol
```diff
@@ -58,97 +122,365 @@ contract KeeperIncentive is ACLAuth, ContractRegistryAccess {
...
+      _burn(burnAmount, incentiveAccountId);
+    }
+  }
+
+  function _previewPayout(Incentive memory incentive) internal pure returns (uint256 payoutAmount, uint256 burnAmount) {
+    burnAmount = (incentive.reward * incentive.burnRate) / 1e18;
```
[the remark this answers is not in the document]
K-G-J: done

packages/hardhat/contracts/core/utils/KeeperIncentive.sol
```diff
@@ -58,97 +122,365 @@ contract KeeperIncentive is ACLAuth, ContractRegistryAccess {
...
+    _cacheKeeperAccount(_keeper, incentiveAccountId);
+
+    incentiveBalances[incentiveAccountId] -= _amount;
+    keeperBalances[_keeper][incentiveAccountId] += _amount;
+  }
+
+  function claim(bytes32[] calldata incentiveAccountIds) external {
+    for (uint256 i; i < incentiveAccountIds.length; i++) {
+      bytes32 incentiveAccountId = incentiveAccountIds[i];
+      _claim(incentiveAccountId, msg.sender);
```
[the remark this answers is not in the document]
K-G-J: done

packages/hardhat/contracts/core/utils/KeeperIncentive.sol
```diff
@@ -58,97 +122,365 @@ contract KeeperIncentive is ACLAuth, ContractRegistryAccess {
...
     uint256 _reward,
     bool _enabled,
-    bool _openToEveryone
-  ) public onlyRole(DAO_ROLE) {
-    incentives[_contractName].push(Incentive({ reward: _reward, enabled: _enabled, openToEveryone: _openToEveryone }));
-    emit IncentiveCreated(_contractName, _reward, _openToEveryone);
+    bool _openToEveryone,
+    address _rewardToken,
+    uint256 _cooldown,
+    uint256 _burnRate
```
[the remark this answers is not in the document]
K-G-J: done

packages/hardhat/contracts/core/utils/KeeperIncentive.sol
K-G-J: done

packages/hardhat/contracts/core/utils/KeeperIncentive.sol
K-G-J: done

packages/hardhat/contracts/core/utils/KeeperIncentive.sol
```diff
@@ -16,32 +16,96 @@ contract KeeperIncentive is ACLAuth, ContractRegistryAccess {
     uint256 reward; //pop reward for calling the function
     bool enabled;
     bool openToEveryone; //can everyone call the function to get the reward or only approved?
+    address rewardToken;
+    uint256 cooldown; // seconds required between incentive payouts
+    uint256 burnRate;
+    uint256 id;
   }
```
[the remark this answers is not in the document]
anthonymartin: agreed - done

packages/hardhat/contracts/core/interfaces/IKeeperIncentiveV1.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: GPL-3.0
+// Docgen-SOLC: 0.8.0
+
+pragma solidity ^0.8.0;
+
+interface IKeeperIncentive {
```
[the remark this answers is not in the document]
anthonymartin: agreed - done

packages/hardhat/contracts/core/utils/KeeperIncentive.sol
```diff
@@ -16,32 +16,96 @@ contract KeeperIncentive is ACLAuth, ContractRegistryAccess {
...
+   */
+  mapping(address => uint256) public keeperClaims;
+
+  /**
+   * pending burn balance
+   */
+  uint256 public burnBalance;
+
+  uint256 public incentivesAccountsCount;
   address internal immutable burnAddress = 0x000000000000000000000000000000000000dEaD; // Burn Address
```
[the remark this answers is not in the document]
anthonymartin: done

packages/hardhat/lib/adapters/KeeperIncentives/getKeeperAccounts.ts
```diff
@@ -0,0 +1,9 @@
+import { KeeperIncentive } from "../../../typechain/KeeperIncentive";
```
anthonymartin, line 1: @K-G-J i think we can delete this

packages/hardhat/lib/adapters/KeeperIncentives/getIncentiveBalance.ts
```diff
@@ -0,0 +1,10 @@
+import { KeeperIncentive } from "../../../typechain/KeeperIncentive";
```
anthonymartin, line 1: @K-G-J   i think we can delete this

packages/hardhat/test/__data__/KeeperIncentives/rewardsTokensIncentiveAccounts.json
```diff
@@ -0,0 +1,8 @@
+[
```
anthonymartin, line 1: @K-G-J i think we can delete this

packages/hardhat/test/__data__/KeeperIncentives/keeperIncentives.ts
```diff
@@ -0,0 +1,144 @@
+export default [
```
anthonymartin, line 1: @K-G-J i think we can delete this

packages/hardhat/lib/adapters/KeeperIncentives/getKeeperAccounts.ts
K-G-J: agreed

packages/hardhat/lib/adapters/KeeperIncentives/getIncentiveBalance.ts
K-G-J: agreed

packages/hardhat/test/__data__/KeeperIncentives/rewardsTokensIncentiveAccounts.json
K-G-J: agreed

packages/hardhat/test/__data__/KeeperIncentives/keeperIncentives.ts
K-G-J: agreed

packages/hardhat/contracts/core/utils/KeeperIncentiveV2.sol
```diff
@@ -0,0 +1,584 @@
...
+    require(
+      block.timestamp - incentives[msg.sender][_i].lastInvocation >= incentive.cooldown,
+      "wait for cooldown period"
+    );
+
+    incentives[msg.sender][_i].lastInvocation = block.timestamp;
+
+    if (!incentive.openToEveryone) {
+      _requireRole(KEEPER_ROLE, _keeper);
+    }
```
ecmendenhall, line 445: Just a nitpick, let's put this check before the effects.

```suggestion
    if (!incentive.openToEveryone) {
      _requireRole(KEEPER_ROLE, _keeper);
    }
    
    incentives[msg.sender][_i].lastInvocation = block.timestamp;
```

packages/hardhat/contracts/core/utils/KeeperIncentiveV2.sol
```diff
@@ -0,0 +1,584 @@
...
+    require(
+      IStaking(_getContract(keccak256("PopLocker"))).balanceOf(_keeper) >= requiredKeeperStake,
+      "not enough pop staked"
+    );
+
+    IKeeperIncentiveV2.Incentive memory incentive = incentives[msg.sender][_i];
+    bytes32 _incentiveAccountId = incentiveAccountId(msg.sender, _i, incentive.rewardToken);
+
+    require(
+      block.timestamp - incentives[msg.sender][_i].lastInvocation >= incentive.cooldown,
```
ecmendenhall, line 437: You're reading this into memory on L#433 and can continue using it:

```suggestion
      block.timestamp - incentive.lastInvocation >= incentive.cooldown,
```

packages/hardhat/contracts/core/utils/KeeperIncentiveV2.sol
```diff
@@ -0,0 +1,584 @@
...
+   * @param _incentiveAccountId id of the incentive
+   * @param tokenAddress address of token to burn
+   */
+  function _burn(
+    uint256 amount,
+    bytes32 _incentiveAccountId,
+    address tokenAddress
+  ) internal {
+    burnBalances[tokenAddress] += amount;
+    _internalTransfer(_incentiveAccountId, address(this), address(0), amount);
```
ecmendenhall, line 552: Would it make sense to use `burnAddress` as the internal account for burned tokens as well?

ecmendenhall (commented):
Leaving some first pass comments here! This is pretty much all style and nitpicks, I will give this another pass for the bigger stuff.

packages/hardhat/contracts/core/interfaces/IKeeperIncentiveV2.sol
```diff
@@ -0,0 +1,47 @@
+// SPDX-License-Identifier: GPL-3.0
+// Docgen-SOLC: 0.8.0
+
+pragma solidity ^0.8.0;
+
+// keccak256("INCENTIVE_MANAGER_ROLE")
+bytes32 constant INCENTIVE_MANAGER_ROLE = 0x3bd2f1998541a778f2b17a84ff608b734af3ddf06cc43506453ef362dcdf17f1;
```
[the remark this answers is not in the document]
K-G-J: good to know!

packages/hardhat/contracts/core/utils/KeeperIncentiveV2.sol
```diff
@@ -0,0 +1,584 @@
...
+  function createIncentive(
+    address _address,
+    uint256 _reward,
+    bool _enabled,
+    bool _openToEveryone,
+    address _rewardToken,
+    uint256 _cooldown,
+    uint256 _burnPercentage
+  ) public onlyRole(INCENTIVE_MANAGER_ROLE) returns (uint256) {
+    require(_cooldown > 0, "must set cooldown");
```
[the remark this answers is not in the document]
K-G-J: @anthonymartin I think @ecmendenhall has a good point here with making a cooldown lower than the average blocktime.  I was thinking we could make it 15 seconds since the average mainnet blocktime is 12 -14 seconds.

packages/hardhat/contracts/core/interfaces/IKeeperIncentiveV2.sol
```diff
@@ -0,0 +1,47 @@
...
+
+  function handleKeeperIncentive(uint8 _i, address _keeper) external;
+
+  function handleKeeperIncentive(
+    bytes32,
+    uint8 _i,
+    address _keeper
+  ) external;
+
+  function tip(
```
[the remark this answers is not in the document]
K-G-J: I think we should default to putting all our functions into our interfaces IMO

packages/hardhat/contracts/core/utils/KeeperIncentiveV2.sol
```diff
@@ -0,0 +1,586 @@
...
+  function _internalTransfer(
+    bytes32 _incentiveAccountId,
+    address from,
+    address to,
+    uint256 amount
+  ) private {
+    IKeeperIncentiveV2.Account storage fromAccount = accounts[_incentiveAccountId][from];
+    IKeeperIncentiveV2.Account storage toAccount = accounts[_incentiveAccountId][to];
+    fromAccount.balance -= from == address(0) ? 0 : amount;
+    toAccount.balance += to == address(0) ? 0 : amount;
```
ecmendenhall, line 538: It might be preferable to treat `address(0)` just like any other account here rather than special casing it. One benefit is that we could test the invariants here—the internal zero address balance should always be the sum of all tokens burned.

ecmendenhall (commented):
Here comes round 2 :smile:

Left a few more comments here. Here is a checklist of some questions I asked and answered, and a few open questions.

**Answered questions:**
(But please check my work!)
- What prevents unauthorized callers from creating an incentive?
    - Creation is protected by ACLAuth and limited to the incentive manager role.
- What prevents unauthorized callers from claiming any incentive?
    - The caller must be an "allowed controller," added to the allowlist when the incentive is created
- Can attackers extract rewards through `claimFor`?
    - No, this always transfers the accrued balance to the keeper. An attacker can pass an invalid keeper address, but it will have a zero balance and revert.
- Do internal accounting invariants work as expected? (Zero transfers, self transfers)
    - Zero transfers are allowed (but change balances by zero)
    - Self transfers are allowed, accounting looks OK
    - Zero address is special-cased in internal transfers. Consider not doing this, so we can test internal invariants.
- What kind of tokens will this contract support?
    - We should avoid weird ERC20s like fee-on-transfer and rebasing tokens as keeper incentives, which could mess with internal accounting. For example, `tip` might transfer fewer tokens than expected if the tip token is a fee-on-transfer. `tip` is probably worse than `fundIncentive` because it will be called with smaller amounts more frequently. (USDT and USDC are both technically optional fee-on-transfer tokens, but it is pretty unlikely they ever turn these on). If we really want to be safe, we could follow a balance before/after pattern in `fundIncentive` and `tip`. 
    
**Questions/potential issues:**
- I see a potential griefing path through the `claimFor` mechanism. An external party can force a keeper to claim accrued rewards, even if unwanted by the keeper. The benefit of this is that someone else can cover the gas costs, but the downside is that it might have tax implications for the keeper. 
- There seems to be no mechanism to migrate deposited incentives in case of emergency or upgrade. What's the game plan if we ever need to replace the contract?
- What is the failure mode for dependent contracts if this one runs out of incentive tokens? Do calls to keeper incentivized functions revert? Can we better defend against failures here?
- A malicious admin could potentially set the burn percentage to 100% and burn all incentive. If we have a malicious admin, we have bigger problems, but it might be worth constraining this to build trust with keepers.

packages/hardhat/test/KeeperIncentiveV2.test.ts
```diff
@@ -0,0 +1,1432 @@
...
+    await mockPop.connect(keeper).approve(staking.address, parseEther("100000"));
+    await rewardToken.connect(manager).approve(keeperIncentive.address, parseEther("100000"));
+    await rewardToken.connect(manager).approve(keeperIncentiveHelper.address, parseEther("100000"));
+    await staking.connect(keeper).lock(keeper.address, parseEther("2000"), 0);
+    await timeTravel(7 * DAYS);
+  };
+
+  beforeEach(async function () {
+    [keeper, manager] = await ethers.getSigners();
+    await setup();
```
ecmendenhall, line 112: Very much a nitpick, but I think the pattern we've adopted [in other tests](https://github.com/popcorndao/sweet-caramel/blob/df6f7c17d4876734c7bf7bdf82da94a238c1652c/packages/hardhat/test/ButterBatchProcessing.test.ts#L190) of returning the deployed contracts from the test `setup()` function is a good one to follow.

packages/hardhat/test/KeeperIncentiveV2.test.ts
```diff
@@ -0,0 +1,1432 @@
...
+          getIncentiveAccountId(keeperIncentiveHelper.address, 2, mockPop.address),
+          getIncentiveAccountId(keeperIncentiveHelper.address, 3, mockPop.address),
+          getIncentiveAccountId(keeperIncentiveHelper.address, 4, mockPop.address),
+          getIncentiveAccountId(keeperIncentiveHelper.address, 5, mockPop.address),
+        ])
+      );
+    });
+
+    it("returns incentive account balances", async () => {
+      expect(false);
```
ecmendenhall, line 182: 🤔 What's going on here? This just does nothing, right?

packages/hardhat/test/KeeperIncentiveV2.test.ts
```diff
@@ -0,0 +1,1432 @@
...
+      );
+
+      await expectValue(
+        JSON.stringify(account0),
+        JSON.stringify({
+          rewardToken: mockPop.address,
+          amount: parseEther("7.5"),
+          account: getIncentiveAccountId(keeperIncentiveHelper.address, 0, mockPop.address),
+        })
+      );
```
ecmendenhall, line 220: Ah, nice pattern here for structs. Maybe this deserves its own `expectStruct` helper.

packages/hardhat/test/KeeperIncentiveV2.test.ts
```diff
@@ -0,0 +1,1432 @@
...
+  });
+});
+
+export async function impersonateAccount(address: string): Promise<SignerWithAddress> {
+  await network.provider.request({
+    method: "hardhat_impersonateAccount",
+    params: [address],
+  });
+  return await ethers.getSigner(address);
+}
```
ecmendenhall, line 1405: This looks like it would be useful as a separate helper.

ecmendenhall (approved):

packages/hardhat/contracts/core/utils/KeeperIncentiveV2.sol
```diff
@@ -0,0 +1,586 @@
...
+    uint256 _amount
+  ) external {
+    require(_amount > 0, "must send amount");
+    require(incentives[_contractAddress].length > _i, "incentive does not exist");
+    IKeeperIncentiveV2.Incentive storage incentive = incentives[_contractAddress][_i];
+
+    bytes32 _incentiveAccountId = incentiveAccountId(_contractAddress, _i, incentive.rewardToken);
+
+    _internalTransfer(_incentiveAccountId, address(0), address(this), _amount);
+    IERC20(incentive.rewardToken).safeTransferFrom(msg.sender, address(this), _amount);
```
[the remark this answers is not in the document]
K-G-J: I think this would be the preferred pattern to help protect against no ERC20 tokens getting transferred in or mismatches between _amount funded and the actual amount transferred @anthonymartin what do you think?

packages/hardhat/contracts/core/utils/KeeperIncentiveV2.sol
K-G-J: I believe this is addressed by now sending burned tokens straight to the burn address balance in the update internal _burn function 
```solidity
  function _burn(
    uint256 amount,
    bytes32 _incentiveAccountId,
    address tokenAddress
  ) internal {
    burnBalancesByToken[tokenAddress] += amount;
    _internalTransfer(_incentiveAccountId, address(this), burnAddress, amount);
  }
```

packages/hardhat/test/KeeperIncentiveV2.test.ts
K-G-J: done 👍

ecmendenhall:
> We debated using the upgrading process from ButterBatchStorage, which still may possibly be implemented if deemed necessary. However, the current game plan uses two interfaces for keepers to claim their rewards. If we upgrade, we will allow keepers to claim all their rewards from the KeeperIncentivesV2 contract using one front-end and another front-end for them to claim their rewards with the new contract. Over time, if all tokens are claimed out of the previous KeeperIncentivesV2 contract, we will deactivate the claiming interface.

This all makes sense for claims. But isn't it possible for incentives to be overfunded? If someone has used `fundIncentive` to cover incentive payments that have not yet been earned by keepers, that amount would be lost, right?

packages/hardhat/test/KeeperIncentiveV2.test.ts
K-G-J: good catch, removed the expect(false)

packages/hardhat/test/KeeperIncentiveV2.test.ts
K-G-J: good idea! We now have an expectStruct helper

K-G-J:
> > We debated using the upgrading process from ButterBatchStorage, which still may possibly be implemented if deemed necessary. However, the current game plan uses two interfaces for keepers to claim their rewards. If we upgrade, we will allow keepers to claim all their rewards from the KeeperIncentivesV2 contract using one front-end and another front-end for them to claim their rewards with the new contract. Over time, if all tokens are claimed out of the previous KeeperIncentivesV2 contract, we will deactivate the claiming interface.
> 
> This all makes sense for claims. But isn't it possible for incentives to be overfunded? If someone has used `fundIncentive` to cover incentive payments that have not yet been earned by keepers, that amount would be lost, right?

Yes so @anthonymartin and I discussed that incentive managers would need to know not to fund incentives for super long periods of time to avoid, but there is probably a better way to handle this. Let me think on it and try some things out.

anthonymartin:
> > > We debated using the upgrading process from ButterBatchStorage, which still may possibly be implemented if deemed necessary. However, the current game plan uses two interfaces for keepers to claim their rewards. If we upgrade, we will allow keepers to claim all their rewards from the KeeperIncentivesV2 contract using one front-end and another front-end for them to claim their rewards with the new contract. Over time, if all tokens are claimed out of the previous KeeperIncentivesV2 contract, we will deactivate the claiming interface.
> > 
> > 
> > This all makes sense for claims. But isn't it possible for incentives to be overfunded? If someone has used `fundIncentive` to cover incentive payments that have not yet been earned by keepers, that amount would be lost, right?
> 
> Yes so @anthonymartin and I discussed that incentive managers would need to know not to fund incentives for super long periods of time to avoid, but there is probably a better way to handle this. Let me think on it and try some things out.

Fund incentives wouldn't be used right away actually, so I don't think it's a big deal.

packages/hardhat/contracts/core/utils/KeeperIncentiveV2.sol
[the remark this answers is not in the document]
anthonymartin: @K-G-J agreed

RedVeil:
@anthonymartin @K-G-J @ecmendenhall Logic looks sound to me. Its a lot simpler than before. But i would like to add some more empty lines in KeeperIncentiveV2 to group logic better and make it easier to read. Especially `tip` is quite the offender there.

PR #730, opened 2022-05-17
Show Dollar Value TVL for stake pages.

bhanusanghi:


packages/app/hooks/useGetPopTokenPriceInUSD.ts
```diff
@@ -0,0 +1,26 @@
...
+  // Get the Pop Price using the uni v3 pool on mainnet
+  const shouldFetch = !!coingeckoClient;
+  return useSWR(
+    shouldFetch ? ["getPopTokenPrice"] : null,
+    async () => getCoinPriceFromCoinGecko(coinGeckoPopTokenApiId),
+    {
+      focusThrottleInterval: 5000,
+    },
+  );
+}
```
arne-z, line 26: We can get the pop price without coingecko, you've done that over in packages/app/hooks/useGetPopUsdcLpTokenPriceInUSD.ts for example. I think it would be good to be consistent and replace everywhere in the codebase where we use the pop price with the same hook util/hook. Then we can worry about wether the hook should internally use 
```
const [usdcAmount, popAmount] = await popUsdcLp.getUnderlyingBalances();
const popPrice = usdcAmount.mul(parseEther("1")).div(popAmount);
```
or whether that should use coingecko.

arne-z:
image
Also we have this problem now

packages/app/components/StakeCard.tsx
```diff
@@ -16,50 +16,53 @@ interface StakeCardProps {
...
-      <div className="flex flex-row items-center justify-between">
-        <div className="flex flex-row items-center">
-          <TokenIcon token={getSanitizedTokenDisplayName(stakedToken?.name)} />
-          <h3 className="secondary-title ml-4 ">{getSanitizedTokenDisplayName(stakedToken?.name)}</h3>
-        </div>
-        <div className="w-24 flex-shrink-0">
-          <MainActionButton
-            label="Stake"
-            handleClick={async () => await onSelectPool(stakingPool?.address, stakedToken?.address)}
-          />
```
arne-z, line 38: Where did the main action button go, and what happened to the styling here.

packages/app/components/StakeCard.tsx
```diff
@@ -16,50 +16,53 @@ interface StakeCardProps {
...
-        </div>
-        <div className="w-1/2 md:w-1/4 mt-4">
-          <StatusWithLabel content={formatStakedAmount(stakingPool.totalStake)} label="Total Staked" />
-        </div>
-        <div className="w-full md:w-1/2 mt-4">
-          <StatusWithLabel
-            content={`${formatAndRoundBigNumber(stakingPool.tokenEmission, 3)} POP / day`}
-            label="Token Emissions"
-          />
+  const dollarPrice = useGetDollarPrice(stakedToken?.address);
```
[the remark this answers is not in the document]
bhanusanghi: With Hooks it is recommended we always fetch the hook instance at a granular level, though it breaks the previously followed conventions of smart containers and dumb components.

packages/app/hooks/useGetDollarPrice.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export default function useGetDollarPrice(token: Address | undefined): BigNumber | undefined {
+  const { data: popPrice, error: err1 } = useGetPopTokenPriceInUSD();
+  const { data: popUsdcLpPrice, error: err2 } = useGetPopUsdcLpTokenPriceInUSD();
+  const { data: butterPrice, error: err3 } = useGetButterTokenPriceInUSD();
+  const { contractAddresses } = useWeb3();
+  popPrice && console.log("popPrice", ethers.utils.parseEther(popPrice.toString()).toString());
+  popUsdcLpPrice &&
+    console.log("popUsdcLpPrice", ethers.utils.parseEther(ethers.utils.formatUnits(popUsdcLpPrice, 6)).toString());
+  butterPrice && console.log("butterPrice", butterPrice.toString());
```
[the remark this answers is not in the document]
bhanusanghi: done

packages/app/hooks/useGetDollarPrice.ts
```diff
@@ -0,0 +1,26 @@
+import { Address } from "@popcorn/utils/types";
+import { BigNumber, ethers } from "ethers";
+import useGetButterTokenPriceInUSD from "hooks/useGetButterTokenPriceInUSD";
+import useGetPopTokenPriceInUSD from "hooks/useGetPopTokenPriceInUSD";
+import useGetPopUsdcLpTokenPriceInUSD from "hooks/useGetPopUsdcLpTokenPriceInUSD";
+import useWeb3 from "./useWeb3";
+
+export default function useGetDollarPrice(token: Address | undefined): BigNumber | undefined {
```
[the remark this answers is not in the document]
bhanusanghi: done

packages/app/hooks/useGetDollarPrice.ts
```diff
@@ -0,0 +1,26 @@
...
+    console.log("popUsdcLpPrice", ethers.utils.parseEther(ethers.utils.formatUnits(popUsdcLpPrice, 6)).toString());
+  butterPrice && console.log("butterPrice", butterPrice.toString());
+  if (!token) return undefined;
+  switch (token.toLowerCase()) {
+    case contractAddresses.pop?.toLowerCase():
+      return popPrice ? ethers.utils.parseEther(popPrice.toString()) : undefined;
+    case contractAddresses.butter?.toLowerCase():
+      return butterPrice;
+    case contractAddresses.popUsdcLp?.toLowerCase():
+      return popUsdcLpPrice ? ethers.utils.parseEther(ethers.utils.formatUnits(popUsdcLpPrice, 6)) : undefined;
```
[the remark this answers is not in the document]
bhanusanghi: With Hooks it is recommended we always fetch the hook instance at a granular level, though it breaks the previously followed conventions of smart containers and dumb components
It would also help in designing a deeplink supported website where you can directly access stake/[id] page without having to first go to the stake page and then clicking on a particular card to access stake/[id].

packages/app/hooks/useGetPopTokenPriceInUSD.ts
bhanusanghi: I am not sure if using the Pop-usdc G-Uni pool is the best way to fetch the PopPrice. 
@anthonymartin what is the configuration for coin gecko to fetch the PopPrice?
Also I have seen that for fetching price using the UniV3 pool we need to integrate their SDK which might unnecessarily bloat our codebase.

While calculating the LP Pool value however, we will need to calculate it this way as the underlying assets are these.
```
const [usdcAmount, popAmount] = await popUsdcLp.getUnderlyingBalances();
const popPrice = usdcAmount.mul(parseEther("1")).div(popAmount);
```

RedVeil: @bhanusanghi why is it not the best way to fetch the pop price? Since there wont be any other LPs but the ones coming from arrakis/gelato we get the whole usdc/pop amount. This allows us to precisly calculate the price of pop. which we already do for getting the apy.

packages/app/components/StakeCard.tsx
bhanusanghi: It looks like it is a change coming from merging with develop branch. Must be the automerge that affected it somehow. Rectified now.

packages/app/components/StakeCard.tsx
```diff
@@ -16,50 +16,53 @@ interface StakeCardProps {
...
+  if (dollarPrice) {
```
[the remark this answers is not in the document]
bhanusanghi: rectified.

packages/app/components/StakeCard.tsx
```diff
@@ -16,50 +16,53 @@ interface StakeCardProps {
...
+                  stakingPool.apy.lt(constants.Zero) ? "New 🍿✨" : formatAndRoundBigNumber(stakingPool.apy, 2) + "%"
+                }
+                label="Est. APY"
+                green
+              />
+            </div>
+            <div className="w-1/2 md:w-1/4 mt-4">
+              <StatusWithLabel
+                content={
+                  formatStakedAmount(stakingPool.totalStake.mul(dollarPrice).div(ethers.constants.WeiPerEther)) + "$"
```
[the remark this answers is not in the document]
bhanusanghi: Done. Since we now show dollar value and not tokens held, I have made the value an int rather than float.

packages/app/hooks/useGetPopTokenPriceInUSD.ts
bhanusanghi: Since the liquidity was not consolidated. If we are planning to consolidate then no issues. Updated the code.

anthonymartin: @RedVeil @bhanusanghi does this return an accurate value for determining pop price if there are LPs outside of the g-uni pool?

why don't we use coingecko price for everything?

RedVeil: "why don't we use coingecko price for everything?" - I was under the impression that you didnt want to rely on coingecko. but its certainly the easiest

RedVeil: about correct pricing id say we do. With two pools that both use G-UNI the large majority of funds should be in the G-UNI contracts so they should hold the overhwelming majority of funds. Currently G-UNI has 480k USDC and 376k POP out of 531k USDC and 386k POP

RedVeil (approved):

PR #757, opened 2022-05-25
Adds Metadata Override features, and uses the contract metadata while displaying.

bhanusanghi:


anthonymartin:
@bhanusanghi could you add token icons here to so they are all in one place?

RedVeil (approved):

PR #761, opened 2022-05-26
factors out batch accounting into it's own contract

anthonymartin:


anthonymartin:
this PR takes the refactor to another level: https://github.com/popcorndao/sweet-caramel/pull/768

PR #785, opened 2022-06-06
decimals required for  all big number conversions

Amirjab21:
The idea is to explicitly require the decimal places whenever we convert from bigNumber to number or vice versa. This is to ensure consistency and correctness.

Sometimes, we will be combining two different tokens e.g input of a 10^6 token and output of a 10^18 token.

e.g user inputs 20 USDC and wants to know how much Butter they get back.

20 USDC is converted to Big Number --> 20 * 10^6

Then, you do whatever calculation is needed. e.g. multiply by 2.

Then, to get the amount of Butter the user gets back, we have to convert it back to a number by passing in 6 for decimal places.


RedVeil:
@Amirjab21 I think we should get rid of formatBigNumber, bigNumberToNumber, numberToBigNumber and scaleNumberToBigNumber in formatBigNumber.ts. Same goes for formatAmounts.ts.

If we want to format smth for display purposes we should use formatAndRoundBigNumber but the version in sweet-pool-integration. https://github.com/popcorndao/sweet-caramel/blob/357dafb58ece2e28124128324eaaffbcb40d5ebf/packages/utils/src/formatBigNumber.ts#L7
This one should require decimals explicitly.

RedVeil:
@Amirjab21 Additionally every token now has decimals so we should be able to use Token.decimals in each of those calls. This means we can get rid of tokenDecimals. Some functions use formatAndRoundBigNumber or similar functions inside. Here we should make sure that they also take a token type as input so we can pass the decimals correctly.

RedVeil:
@Amirjab21 We have quite a bit conversions in the ButterInput. I can take a closer look at this afterwards to make sure we dont convert wrong here. I remember multiple conversions in that whole flow. Most importantly we should not raise any values to 1e18 if they token doesnt use that many decimals.

Amirjab21:
@RedVeil Are you sure we can get rid of bigNumberToNumber? What if we need to do some calculations on the front-end? Wouldn't the rounding affect these calculations and therefore require us to use bigNumberToNumber?

Amirjab21:
I got rid of everything to only keep the remaining

formatAndRoundBigNumber(value, decimals)

numberToBigNumber(value, decimals)

formatStakedAmount still exists but it uses formatAndRoundBigNumber under the hood, same with formatStakedTVL

PR #800, opened 2022-06-09
Feature/transaction notifications

Amirjab21:


packages/app/components/Notifications/ToastContainer.tsx
```diff
@@ -0,0 +1,63 @@
+import { bigNumberToNumber } from "@popcorn/utils";
+import toast from "react-hot-toast";
+
+export enum methodNames {
```
RedVeil, line 4: I think this should be more explicit. Maybe `ToastMethod`.

packages/app/components/Notifications/ToastContainer.tsx
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  return toast[messageType](title, options);
+}
+
+const customNotification = (methodName: methodNames, res: any, messageType: messageType, title: string, args: any) => {
+  switch (methodName) {
+    case methodNames.ZapIntoBatch:
+      const eventLog = res.events.filter((event) => event.event === "ZappedIntoBatch").args;
+      return toast[messageType](
```
RedVeil, line 47: I think we have to use toast.custom() to display a toast that fully looks like the design.

RedVeil: (With changed border radius etc. You can find docs for it on their site. It works perfect with tailwind)

packages/app/components/Notifications/ToastContainer.tsx
```diff
@@ -0,0 +1,63 @@
+import { bigNumberToNumber } from "@popcorn/utils";
+import toast from "react-hot-toast";
+
+export enum methodNames {
+  ZapIntoBatch = "ZapIntoBatch",
+}
+
+export enum messageType {
```
RedVeil, line 8: I think we should make these also `ToastMethods`. We have to customize all calls now anyways so we can also spend the time and go into each function call to set these things explicitly

packages/app/components/Notifications/ToastContainer.tsx
```diff
@@ -0,0 +1,63 @@
...
+  success = "success",
+  error = "error",
+  loading = "loading",
+  transactionSuccess = "success",
+}
+
+const options = {};
+
+export default function ToastContainer(
```
RedVeil, line 17: Id like to rename this to `showToast` or smth like that. Its a bit easier to understand in the functions that call it. The original syntax is great `toast.success()`! Id like to get close to smth like that.

packages/app/components/Notifications/ToastContainer.tsx
```diff
@@ -0,0 +1,63 @@
...
+}
+
+export enum messageType {
+  success = "success",
+  error = "error",
+  loading = "loading",
+  transactionSuccess = "success",
+}
+
+const options = {};
```
RedVeil, line 15: What are options for?

packages/app/components/Notifications/ToastContainer.tsx
```diff
@@ -0,0 +1,63 @@
...
+  messageType: messageType,
+  res?: any,
+  methodName?: methodNames,
+  args?: any,
+  description?: string,
+) {
+  toast.dismiss();
+  if (methodName) {
+    return customNotification(methodName, res, messageType, title, args);
+  } else if (description) {
```
RedVeil, line 28: When we use only `MethodType` we should have a base toast container which just swaps out the text and icon via a switch case and a customNotification as you did here which again uses a switch case for more specialized designs

packages/app/pages/[network]/butter/index.tsx
```diff
@@ -238,31 +238,41 @@ export default function Butter(): JSX.Element {
...
-        },
-        "hideMintPopover",
-        dispatch,
-      );
-    });
+          "hideMintPopover",
+          dispatch,
+        );
+      },
+      methodNames.ZapIntoBatch,
```
Comment (RedVeil) [inline packages/app/pages/[network]/butter/index.tsx:269]: This is not zapping into the batch but only depositing into the current batch

packages/app/components/Notifications/ToastContainer.tsx
RedVeil: We would need to decide maybe with @FanFanDAO and @anthonymartin how many different methods we want to have.

packages/app/components/Notifications/ToastContainer.tsx
Amirjab21: removed

packages/app/components/Notifications/ToastContainer.tsx
Amirjab21: Theres a problem with the toast.custom() - the animation just doesn't work properly.

Look at this issue: https://github.com/timolins/react-hot-toast/issues/116

I've put it on the latest PR but it doesn't work as others have also complained about.

packages/app/pages/[network]/butter/index.tsx
Comment (Amirjab21) [reply packages/app/pages/[network]/butter/index.tsx]: done

packages/app/components/Notifications/ToastContainer.tsx
Amirjab21: Think I've implemented it - check it out

RedVeil:
@Amirjab21 Great PR! Thinking more about it though i think we should get rid of custom notifications. All notifications should be custom either of type loading, success or error. Error works perfectly as is but we should simply input a <p> with the custom information in loading and success.

RedVeil:
@Amirjab21 Additionally i think its time that onContractSuccess gets an input object with all the params. its getting kinda hard to understand when adding input params what is required and what they actually are there for.

RedVeil:
@Amirjab21 Also can we rename ShowToast to TransactionToast?

RedVeil:
i transfered this code into a new PR since it was easier than merging this into new-website

PR #804, opened 2022-06-09
adds mint/redeem fee

anthonymartin:
* adds mint fee and collects fee in batch mint function
* moves redeem fee to batch redeem function
* removes sweethearts
* adds tests


packages/hardhat/contracts/core/defi/three-x/ThreeXBatchProcessing.sol
```diff
@@ -344,17 +356,21 @@ contract ThreeXBatchProcessing is ACLAuth, KeeperIncentivized, AbstractBatchCont
       );
     }
 
-    uint256 claimableTokenBalance = batch.targetToken.balanceOf(address(this));
+    uint256 claimableTokenBalance = batch.targetToken.balanceOf(address(this)) - redemptionFee.accumulated;
```
RedVeil, line 359: Could you not just use `setTokenAmountLessFees ` here?

RedVeil, line 359: Could you not just use `setTokenAmountLessFees ` here?

packages/hardhat/contracts/core/defi/three-x/controller/AbstractMintFee.sol
```diff
@@ -0,0 +1,63 @@
...
+pragma solidity ^0.8.0;
+
+import { IERC20, SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "../../../utils/ACLAuth.sol";
+import "@openzeppelin/contracts/utils/math/Math.sol";
+
+abstract contract AbstractMintFee is ACLAuth {
+  using SafeERC20 for IERC20;
+
+  struct MintFee {
```
RedVeil, line 13: I think we dont need two abstracts. All functions are the same. The only thing thats different is calling it mint or redeem. Couldnt we have a mapping bytes32 -> Fee? Bytes32 beeing "mint", "fee", "withdraw" ...

Than we can send the key to _takeFee() function aswell which simply gets the right struct. Same would go for the setter.

RedVeil, line 13: I think we dont need two abstracts. All functions are the same. The only thing thats different is calling it mint or redeem. Couldnt we have a mapping bytes32 -> Fee? Bytes32 beeing "mint", "fee", "withdraw" ...

Than we can send the key to _takeFee() function aswell which simply gets the right struct. Same would go for the setter.

PR #814, opened 2022-06-10
feature: add staking popover

Ajiva-D:


packages/grants/pages/applications/[id].tsx
```diff
@@ -0,0 +1,203 @@
...
+            onCloseStakeModal={closeStakeModal}
+            hasExpired={expired}
+          />
+        ),
+        visible: true,
+      }),
+    );
+  };
+
+  const getLockedPop = async () => {
```
Comment (arne-z) [inline packages/grants/pages/applications/[id].tsx:91]: I would like to rename this, since getLockedPop implies a getter function, whereas this function returns nothing and has side effects. Perhaps updateLockedPopState or something.

Comment (arne-z) [reply packages/grants/pages/applications/[id].tsx:91]: Or refetchLockedPop

arne-z (changes requested):

arne-z (approved):

PR #837, opened 2022-06-14
Feature/beneficiary applications card

Ajiva-D:


packages/app/helper/modalHelpers.ts
```diff
@@ -16,7 +16,7 @@ export const toggleModal = (
   dispatch: React.Dispatch<any>,
 ) => {
   if (!localStorage.getItem(key)) {
-    if (modalType === ModalType.SingleAction) {
+    if ((modalType = ModalType.SingleAction)) {
```
arne-z, line 19: This file should not be included in this PR, revert please.

arne-z (changes requested):

packages/app/helper/modalHelpers.ts
Ajiva-D: Done

arne-z (approved):

PR #856, opened 2022-06-20
Feature/beneficiary applications landing v2

Ajiva-D:


packages/grants/components/CommonComponents/CardBody.tsx
```diff
@@ -1,41 +1,71 @@
...
+    setProgress(Math.round((progressVotes + Number.EPSILON) * 100) / 100);
+  }, [votes]);
   return (
     <div className="shadow-custom-lg bg-white rounded-4xl">
       <div className="relative">
         <img
           className="h-48 w-full object-cover rounded-t-4xl"
-          src={image?.image}
-          alt={image?.description || `Picture of ${name}`}
+          src={`https://popcorn.mypinata.cloud/ipfs/${image?.image}`}
```
tnguyen42, line 46: Isn't there some place where you can store this base url instead of hard-coding it?

packages/grants/pages/applications/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { ViewGridIcon } from "@heroicons/react/outline";
+import { ChevronDownIcon } from "@heroicons/react/solid";
+import { BeneficiaryGovernanceAdapter, Proposal, ProposalStatus, ProposalType } from "@popcorn/hardhat/lib/adapters";
+import { IpfsClient } from "@popcorn/utils";
+import BeneficiaryFilter from "components/Beneficiaries/BeneficiaryFilter";
+import CardBody from "components/CommonComponents/CardBody";
+import { CardLoader } from "components/CommonComponents/CardLoader";
+import { useContext, useEffect, useState } from "react";
+import toast from "react-hot-toast";
+import { ContractsContext } from "../../context/Web3/contracts";
```
tnguyen42, line 11: For readability purposes, I think an absolute import would be better here

packages/grants/components/CommonComponents/CardBody.tsx
```diff
@@ -1,41 +1,71 @@
...
+    setProgress(Math.round((progressVotes + Number.EPSILON) * 100) / 100);
+  }, [votes]);
   return (
     <div className="shadow-custom-lg bg-white rounded-4xl">
       <div className="relative">
         <img
           className="h-48 w-full object-cover rounded-t-4xl"
-          src={image?.image}
-          alt={image?.description || `Picture of ${name}`}
+          src={`${process.env.IPFS_URL}${image?.image}`}
```
arne-z, line 46: The IPFS_URL set in netlify.toml seems to already be correct, I think you should add the env variables you use to the env.example file though.

PR #905, opened 2022-06-28
Fix/voting card fixes

Ajiva-D:


packages/grants/components/Proposals/VotingCard.tsx
```diff
@@ -81,7 +83,8 @@ const VotingCard: React.FC<VotingCardProps> = ({
         <VotingProgress progress={progress} labels={["Yes", "No"]} />
       </Transition>
 
-      {hasStaked ? (
+      {/* hasStaked */}
+      {false ? (
```
arne-z, line 87: this can't be right.

arne-z (changes requested):
There appears to be one ternary commented out for debugging or something, Looks good other than that.

packages/grants/components/Proposals/VotingCard.tsx
Ajiva-D: Ah, I was testing.
Forgot to revert

arne-z (approved):

PR #860, opened 2022-06-21
Feature/instant three x contract

arne-z:

PR #910, opened 2022-06-30
ThreeXWhaleProcessing Contract and Tests (from #860)

arne-z:


arne-z:
This PR Contains isolated changes from #860.

packages/hardhat/contracts/core/defi/three-x/ThreeXWhaleProcessing.sol
```diff
@@ -0,0 +1,486 @@
...
+    IContractRegistry __contractRegistry,
+    BasicIssuanceModule _basicIssuanceModule,
+    IStaking _staking,
+    Curve3Pool _threePool,
+    IERC20[3] memory _token
+  ) AbstractBatchController(_contractRegistry) ContractRegistryAccess(__contractRegistry) {
+    IThreeXBatchProcessing threeXBatchProcessing = IThreeXBatchProcessing(
+      _getContract(keccak256("ThreeXBatchProcessing"))
+    );
+    staking = _staking;
```
RedVeil, line 56: Nitpicky but could we order these like the state variables above and the mint/redeemBatchTokens last?

packages/hardhat/contracts/core/defi/three-x/ThreeXWhaleProcessing.sol
```diff
@@ -0,0 +1,486 @@
...
+  /* ========== CONSTRUCTOR ========== */
+
+  constructor(
+    IContractRegistry __contractRegistry,
+    BasicIssuanceModule _basicIssuanceModule,
+    IStaking _staking,
+    Curve3Pool _threePool,
+    IERC20[3] memory _token
+  ) AbstractBatchController(_contractRegistry) ContractRegistryAccess(__contractRegistry) {
+    IThreeXBatchProcessing threeXBatchProcessing = IThreeXBatchProcessing(
```
RedVeil, line 53: Could you move this above the mintBatchTokens call? I read it at first as a state setter until i realized we only use it for the calls here

packages/hardhat/contracts/core/defi/three-x/ThreeXWhaleProcessing.sol
```diff
@@ -0,0 +1,486 @@
...
+  }
+
+  function _runPersonalMint(
+    uint256 _amount,
+    uint256 _minMintAmount,
+    IThreeXBatchProcessing _threeXBatchProcessing
+  ) internal returns (uint256 setTokenAmountAfterFees) {
+    // Remaining amount of USDC in this batch which hasnt been allocated yet
+    uint256 remainingBatchBalanceValue = _amount;
+    (address[] memory tokenAddresses, uint256[] memory quantities) = basicIssuanceModule
```
RedVeil, line 201: you can pass this from _mint to save that call

packages/hardhat/contracts/core/defi/three-x/ThreeXWhaleProcessing.sol
```diff
@@ -0,0 +1,486 @@
...
+      setTokenAmount - _minMintAmount,
+      setTokenAmount,
+      mintBatchTokens.targetToken
+    );
+
+    //Mint 3X
+    basicIssuanceModule.issue(ISetToken(address(mintBatchTokens.targetToken)), setTokenAmount, address(this));
+  }
+
+  function _mint(
```
RedVeil, line 261: Could you move this function above _runBatchDepositMintClaim and _runPersonalMint?

packages/hardhat/contracts/core/defi/three-x/ThreeXBatchProcessing.sol
```diff
@@ -131,7 +137,7 @@ contract ThreeXBatchProcessing is ACLAuth, KeeperIncentivized, AbstractBatchCont
       uint256 lpTokenPriceInUSD = i == 0
         ? componentDependencies[_tokenAddresses[i]].curveMetaPool.get_virtual_price()
         : (componentDependencies[_tokenAddresses[i]].curveMetaPool.get_virtual_price() *
-          componentDependencies[_tokenAddresses[i]].utilityPool.price_oracle()) / 1e18;
+          (2e18 - componentDependencies[_tokenAddresses[i]].oracle.read())) / 1e18;
```
RedVeil, line 140: I dont understand why it shows a diff here. This is exactly how we handle it on develop.

packages/hardhat/contracts/core/defi/three-x/ThreeXWhaleProcessing.sol
```diff
@@ -0,0 +1,486 @@
...
+  ) internal returns (uint256 claimableTokenBalanceAfterFees) {
+    //Get tokenAddresses for mapping of underlying
+    (address[] memory tokenAddresses, uint256[] memory quantities) = basicIssuanceModule
+      .getRequiredComponentUnitsForIssue(ISetToken(address(redeemBatchTokens.sourceToken)), _amount);
+    uint256 setValue = _threeXBatchProcessing.valueOfComponents(tokenAddresses, quantities);
+    if (
+      _minRedeemAmount >=
+      _threeXBatchProcessing.getMinAmountFromRedeem(setValue, _threeXBatchProcessing.slippage().redeemBps)
+    ) {
+      claimableTokenBalanceAfterFees = _runDepositRedeemClaim(_amount, _threeXBatchProcessing);
```
RedVeil, line 348: Could you add the same comments as you did in _mint?

packages/hardhat/contracts/core/defi/three-x/ThreeXWhaleProcessing.sol
```diff
@@ -0,0 +1,486 @@
...
+        _minRedeemAmount,
+        _threeXBatchProcessing,
+        tokenAddresses
+      );
+    }
+
+    emit Redeemed(_amount, claimableTokenBalanceAfterFees);
+  }
+
+  function _setBasicIssuanceModuleAllowance(uint256 _amount) internal {
```
RedVeil, line 361: Lets move this next to the other approval function

packages/hardhat/contracts/core/defi/three-x/ThreeXWhaleProcessing.sol
```diff
@@ -0,0 +1,486 @@
...
+        swapToken.safeApprove(address(curveMetapool), type(uint256).max);
+      }
+
+      lpToken.safeApprove(yToken[i], 0);
+      lpToken.safeApprove(yToken[i], type(uint256).max);
+
+      lpToken.safeApprove(address(curveMetapool), 0);
+      lpToken.safeApprove(address(curveMetapool), type(uint256).max);
+    }
+    IERC20(threePool.coins(0)).safeApprove(address(threePool), 0);
```
RedVeil, line 158: I think a loop would be cleaner here

RedVeil:
The contract looks good as far as i can tell but i think we should move some functions around and add more spaces in between lines to group certain logic and make it easier to read.

packages/hardhat/test/forge/ThreeXWhaleProcessing.t.sol
```diff
@@ -0,0 +1,333 @@
...
+address constant THREE_POOL = 0xbEbc44782C7dB0a1A60Cb6fe97d0b483032FF1C7;
+address constant THREEX_VAULT = 0x0B4E13D8019D0F762377570000D9C923f0dad82B;
+address constant ACL_REGISTRY = 0x8A41aAa4B467ea545DDDc5759cE3D35984F093f4;
+address constant THREEX_BATCH = 0x42189f909e1EFA409A4509070dDBc31A592422A8;
+address constant DAO = 0x92a1cB552d0e177f3A135B4c87A4160C8f2a485f;
+address constant Y_SUSD = 0x5a770DbD3Ee6bAF2802D29a901Ef11501C44797A;
+address constant Y_THREE_EUR = 0x5AB64C599FcC59f0f2726A300b03166A395578Da;
+
+// Run with block number 15008113
+// forge test --fork-url https://eth-mainnet.alchemyapi.io/v2/KsuP431uPWKR3KFb-K_0MT1jcwpUnjAg --fork-block-number 15008113 --match-contract ThreeXWhaleProcessing -vvv
```
RedVeil, line 36:
```suggestion
// forge test --fork-url https://eth-mainnet.alchemyapi.io/v2/privKey  --fork-block-number 15008113 --match-contract ThreeXWhaleProcessing -vvv
```

packages/hardhat/test/forge/ThreeXWhaleProcessing.t.sol
```diff
@@ -0,0 +1,333 @@
...
+        threeXBatchProcessing.valueOfComponents(tokenAddresses, quantities),
+        _slippageBps
+      );
+  }
+
+  // What to test:
+  function test_mint_high_slippage() public {
+    // Should use personal mint
+    uint256 usdcAmountToMint = 100e6;
+    bytes32 mintbatchId = threeXBatchProcessing.currentMintBatchId();
```
RedVeil, line 134: You could call all mintBatchId,threeXBalanceBefore and usdcBalanceBefore in the setUp function since setUp runs everytime before a test. This way we have less code duplication

packages/hardhat/test/forge/ThreeXWhaleProcessing.t.sol
```diff
@@ -0,0 +1,333 @@
...
+      threeXBatchProcessing.getMinAmountFromRedeem(
+        threeXBatchProcessing.valueOfComponents(tokenAddresses, quantities),
+        _slippageBps
+      );
+  }
+
+  // What to test:
+  function test_mint_high_slippage() public {
+    // Should use personal mint
+    uint256 usdcAmountToMint = 100e6;
```
RedVeil, line 133: Seems like u are using the same amounts here so lets set it as an immutable in state

RedVeil:
The test and its test cases look good to me. I think we should also add some more spaces in between lines there to make it a bit easier to read though. I really like the use of comments to describe expected behavior.
We should probably also think of some attack vectors and test those since its currently just happy case testing.

RedVeil (approved):

PR #932, opened 2022-07-06
Modals fixed

Amirjab21:
modals click off now closes modal and heirarchy introduced between network switch and restaking modal


anthonymartin:
@Amirjab21 some modals - like the sign message modal should not be dismissible.  on desktop, that modal is also showing up at the bottom of the page as it should on mobile. it shouldn't be displayed that way on desktop, so I think some mobile styles are being applied on desktop

Amirjab21:
@anthonymartin I've noticed a new issue. We currently don't have a proper heirarchy for modals and there are other situations where multiple modals show up. My current solution of just having an if condition that stops the re-stake modal showing up if the network switch modal is visible may not be sustainable as a solution.... so should we build some sort of hierarchy or stacked list that only shows one modal at a time?

anthonymartin (approved):

PR #961, opened 2022-07-12
Feature/make 3x and butter permissionless

Amirjab21:
We still need to decide where the cron-job will live. Are we going to add a server to our Nextjs app? What architecture preferences do we have? I can make a custom server at the root of the /app folder like this example:
https://github.com/vercel/next.js/tree/canary/examples/custom-server-express

Also, something with the DAO_ROLE needs to actually run the hardhat task that changes the approval required for these selected incentives. I've provided two example bash scripts so this can be done very quickly.

When we've decided with the server architecture, I can just create an express server and schedule cron-jobs using an npm package like https://www.npmjs.com/package/cron


packages/hardhat/lib/tasks/keeper-incentive/toggleApproval.ts
```diff
@@ -0,0 +1,37 @@
+import { utils } from "ethers";
+import { ethers } from "hardhat";
+import { task } from "hardhat/config";
+import { Address } from "packages/utils/src/types";
+
+interface Args {
+  contract: string;
+  keeperincentiveaddress: Address;
```
RedVeil, line 8:
```suggestion
  keeperincentiveAddress: Address;
```

packages/hardhat/lib/tasks/keeper-incentive/toggleApproval.ts
```diff
@@ -0,0 +1,37 @@
...
+  index: string;
+}
+
+export default task("keeper-incentive:toggleApproval", "updates keeper incentive")
+  .addParam("contract", "name of contract")
+  .addParam("keeperincentiveaddress", "address of keeper incentive")
```
RedVeil, line 14:
```suggestion
  .addParam("keeperIncentiveAddress", "address of keeper incentive")
```

packages/hardhat/lib/tasks/keeper-incentive/toggleApproval.ts
```diff
@@ -0,0 +1,37 @@
...
+  .addParam("index", "index of incentive")
+  .setAction(async (args: Args, hre) => {
+    const signer = hre.ethers.provider.getSigner();
+    const keeperIncentive = await hre.ethers.getContractAt("KeeperIncentive", args.keeperincentiveaddress, signer);
```
RedVeil, line 18:
```suggestion
    const keeperIncentive = await hre.ethers.getContractAt("KeeperIncentive", args.keeperIncentiveAddress, signer);
```

packages/hardhat/lib/utils/getNamedAccounts.ts
```diff
@@ -89,6 +89,12 @@ export default function getNamedAccounts() {
       rinkeby: "0xB4AdA607B9d6b2c9Ee07A275e9616B84AC560139",
       hardhat: "0xB4AdA607B9d6b2c9Ee07A275e9616B84AC560139",
     },
+    yMim: {
```
RedVeil, line 92: the Mim references can get deleted again

packages/hardhat/lib/utils/getNamedAccounts.ts
```diff
@@ -147,6 +153,11 @@ export default function getNamedAccounts() {
       rinkeby: "0xd632f22692FaC7611d2AA1C0D552930D43CAEd3B",
       hardhat: "0xd632f22692FaC7611d2AA1C0D552930D43CAEd3B",
     },
+    crvMimMetapool: {
```
RedVeil, line 156: same here

RedVeil:
I think we should make the getContract function a utility, aswell as the batchMint/batchRedeem. This would allow us to only change the butter/threeX contract in the task. Both tasks (butter/threeX) would than only be differantiated by the task name and contract address.

Amirjab21:
> I think we should make the getContract function a utility, aswell as the batchMint/batchRedeem. This would allow us to only change the butter/threeX contract in the task. Both tasks (butter/threeX) would than only be differantiated by the task name and contract address.

done - check it outttt

anthonymartin (approved):

RedVeil (approved):

PR #981, opened 2022-07-25
exchanged img for next/Image

RedVeil:


packages/app/components/BatchButter/SelectToken.tsx
```diff
@@ -18,14 +19,21 @@ export default function SelectToken({
   selectToken,
 }: SelectTokenProps): JSX.Element {
   const [showDropdown, setDropdown] = useState<boolean>(false);
-
+  console.log(`/images/tokens/${selectedToken.img}`)
```
anthonymartin, line 22:
```suggestion
```

anthonymartin (changes requested):
shouldn't the priority property be used for preloading images?

RedVeil:
> shouldn't the priority property be used for preloading images?

This is still super wip. It takes a little longer than expected to swap out img with nextImages. So currently im still swapping them out and making sure everything looks as expected. After that i check priority and placeholder images

RedVeil:
no longer applicable

PR #1000, opened 2022-07-26
minimum deposit size added to butter, 3x and instantmint

Amirjab21:


RedVeil:
@Amirjab21 We have a hook the get the 3CRV virtual price. So we can work in dollar terms.

RedVeil:
@Amirjab21 Additionally in 3x a user might input usdc/dai or usdt. usdc and usdt have 6 decimals so we cant just raise it to 1e18. dai can be raised as normal. Assuming that these token have a value of 1$ seems fine to me though

RedVeil:
Closed since a new merged ticket solves this

PR #1011, opened 2022-07-29
Add Gnosis support

arne-z:


ecmendenhall:
Fuzz tests found a counterexample here that reverts for a very large deposit (> `995_939` 3crypto LP tokens). It looks like the actual vault deposit limit is a little lower than the max value in the tests.

It would be ideal to look up this value in the test itself by asking Yearn for the current limit, but for now we have set the max a little lower.

arne-z:
These changes were part of https://github.com/popcorndao/sweet-caramel/pull/1013

PR #1078, opened 2022-08-08
New 3x/butter page

Ajiva-D:


RedVeil:
Could we center these icons? 
image

It looks okish for USDC but especially 3X becomes a bit strange. (This is also the case on desktop)

RedVeil:
Redeem has a weird display issue on desktop:
image

RedVeil:
For some reason the image doesnt cover the full width:
image

RedVeil:
Nothing is displayed when i have no batches. I think we should atleast show some placeholder
image

RedVeil:
The Tab symbols are inverted. Instead of the active tab beeing white and the rest having lower opacity the active one has lower opacity.
image

RedVeil:
Step 3 is smaller than the other two:
image

RedVeil:
The Disconnect Button of the Navbar is a bit wider than the content of the page. Is that intended?
image

Ajiva-D:
> Step 3 is smaller than the other two: image

Do you mean the content?

RedVeil:
Even though its not explicitly mentioned in the figma design i think the batches should align the with mintInterface on the side. If someone has more than these batches they should become scrollable.

RedVeil:
> > Step 3 is smaller than the other two: image
> 
> Do you mean the content?

The whole card is smaller. So there is more space all of a sudden between it and the batchProgress below. I think the card size shouldnt change. Just means that the info text in the tutorial will have more space around it.

RedVeil (approved):

PR #1079, opened 2022-08-08
New staking page

Ajiva-D:


RedVeil:
Could we center theses icons? 
image

RedVeil:
Token Emission is in normal case in figma.
image

RedVeil:
Could you rename this to "view" ? 
image

RedVeil:
This image should be full width.
image

Ajiva-D:
> Could we center theses icons? image

I don't know why these look off for you, cos it's actually centralised, maybe try a different browser and see if it's the same so we can pinpoint the problem

RedVeil (approved):

PR #1116, opened 2022-08-16
Add serverless function to check user address vs ofac list.

arne-z:


packages/app/pages/api/checkOfac.ts
```diff
@@ -0,0 +1,19 @@
...
+
+const OFAC_LIST_URL = "https://www.treasury.gov/ofac/downloads/sdnlist.txt";
+
+export default async function handler(req: NextApiRequest, res: NextApiResponse) {
+  const { address } = req.query;
+
+  let response;
+  try {
+    const list: string = (await axios.get(OFAC_LIST_URL)).data;
+    response = typeof address === "string" ? !list.includes(address) : !address.some((entry) => list.includes(entry));
```
anthonymartin, line 12: i think this should `toLowerCase()` before making a comparison

packages/app/components/OfacCheck.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+export default function OfacCheck(): JSX.Element {
+  const { dispatch } = useContext(store);
+  const { account, disconnect } = useWeb3();
+  const [data, setData] = useState<{ success: boolean; permitted: boolean }>(null);
+  const [isLoading, setLoading] = useState(false);
+
+  useEffect(() => {
+    setLoading(true);
+    fetch(`/api/checkOfac?address=${account}`)
```
anthonymartin, line 14: should this be wrapped w/ `if (account) {` ?

arne-z: :+1:

RedVeil (approved):

PR #1125, opened 2022-08-18
docgen fixed by adding OZ/solidity-docgen to repo

Amirjab21:


anthonymartin:
closing since this does not fix the issue

PR #1146, opened 2022-08-24
Feat/deploy staking vault factory

RedVeil:


packages/hardhat/contracts/core/defi/vault/VaultsV1Factory.sol
```diff
@@ -0,0 +1,93 @@
...
+    });
+
+    Staking staking = new Staking(
+      pop,
+      IERC20(address(vault)),
+      IRewardsEscrow(IContractRegistry(_vaultParams.contractRegistry).getContract(keccak256("RewardsEscrow")))
+    );
+    vault.setStaking(address(staking));
+
+    emit VaultV1Deployment(address(vault));
```
K-G-J, line 90: perhaps we may want to update this event to ```VaultDeployment(address(vault), address(staking));``` for the front-end?

RedVeil: yeah that makes sense

PR #1148, opened 2022-08-25
Feat/zap withdraw and unstake zero x zapper

arne-z:


RedVeil:
Could you add a test in Vault.test.ts that makes sure that we dont check allowance if the unstakeAndRedeem function is called by zapper? (It should be similar to the test in Staking where we make sure that this happens when called by vault)

RedVeil (approved):

PR #1166, opened 2022-08-30
Add Zapper for swapping non curve assets and tests.

arne-z:


ecmendenhall:
Made a few updates here @arne-z:

- Renamed `ZeroXZapper` `VaultsV1Zapper`, which is a better description of what it is.
- Renamed `SwapZapper` `ZeroXSwapZapIn`, now that there is not another thing named `ZeroXZapper`. I think this is also a better description of what it actually does.
- Updated the `VaultsV1Zapper` to store a `zapIn` and `zapOut` address for each vault in the `VaultConfig` struct, rather than use the Curve zappers for all vaults. As long as a zapper conforms to the `IZapIn` or `IZapOut` interface it will work here. 

Still to do:
- We need a `ZeroXSwapZapOut` that conforms to `IZapOut` and does the same thing as `ZeroXSwapZapIn` in reverse. (Just like we did with the zap in, we can start with the Zapper Curve ZapOut.
- We should test stargate zaps in and out through the `VaultsV1Zapper`.

RedVeil:
@arne-z @ecmendenhall Could you guys also update Zapper.ts to choose one of the two zappers for an operation? I think we should check if stableSwapAddress == address(0) which would mean we use the swapZapper or otherwise choose the VaultsV1Zapper. Also i think we should rename VaultsV1Zapper to CurveLpZapper or smth like that

ecmendenhall:
@arne-z I've pushed up some changes to store the `zapIn` and `zapOut` address in `VaultsV1Registry` and read them from there in `VaultsV1Zapper` instead of from external storage.

Unfortunately, this was a bigger change than I expected:
- `VaultsV1Zapper#zapOut` is using the max number of local variables internally. (I inlined `_getZapOut` in this function to make it work).
- Similarly, `VaultsV1Factory#deployVaultV1` can't take two more arguments without maxing out the stack, so I had to roll them up into a `VaultParams` struct. 
- Changes to the deployment args and `VaultMetadata` struct cascaded to `VaultsV1Controller` and `VaultsV1Registry.`
- The `VaultsV1Zapper` is back to just storing a mapping of vault asset to vault address.

There are a couple Forge tests failing that I still need to work out: 
- `VaultsV1Registry.t.sol#test__view__getVaultAddressNotRegisteredReverts`
- `VaultsV1Factory.t.sol#test__deployVaultV1NotOwnerReverts`

Lastly, the tests in `ZapperModule.test.ts` will need to set up a `VaultsV1Registry` as part of their test setup in order to use the zapper.

Sorry for leaving you a bit of a mess—if you don't want to deal with this until I'm back on Tuesday, no worries.

RedVeil:
Closed since Stargate has all the new features fomr this PR

PR #1170, opened 2022-09-01
Feat/min deposits for new website

arne-z:


packages/app/helper/isDepositDisabled.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const tvlLimit = parseEther("1000000"); // 1m
+  if (!pageState.redeeming && pageState?.depositAmount.add(tvl).gte(tvlLimit)) {
+    return { disabled: true, errorMessage: "*Exceeds TVL-Limit" };
+  }
+  if (
+    !pageState.redeeming &&
+    !pageState.depositAmount.isZero() &&
+    pageState.depositAmount
+      .mul(pageState.tokens[pageState.selectedToken.input].price)
+      .div(parseUnits("1", 18))
```
Amirjab21, line 23: Doesn't this need to use the token decimals for the input token?

arne-z: It's a question of what decimals we want to end up in when we compare. Currently, I'm multiplying with the price (1e18) and then dividing by 1e18 to get back to the decimals the coin had. Then I compare with the value 100 in whatever decimals the coin had.
It would be possible to instead divide by the coin's decimals here, which is going to end up with a number in 1e18, and then in the next step simply compare with 100*1e18.
I don't have much of a preference here, either one should work.

RedVeil:
@arne-z Code is looking good. Interaction feels smooth. I like it.

RedVeil (approved):

PR #1212, opened 2022-09-08
Feat/grants page

RedVeil:


anthonymartin:
@Amirjab21 - there shouldn't be any diffs any package other than grants in this PR. something seems off - maybe it's formatting?

Amirjab21:
@anthonymartin  I've looked through the files, its all formatting

Except there are some changes to hardhat and utils that were made for grants that should be there.

PR #1340, opened 2022-10-14
connect wallet smart/connected component created and utilised everywh…

Amirjab21:
…ere on the app


arne-z (approved):

arne-z:
A style suggestiong:
For components with boolean props, you can instead of writing:
```
          <ConnectWallet withinPage={true} />
          <ConnectWallet withinPage={false} />
```
write:
```
          <ConnectWallet withinPage />
          <ConnectWallet />
```
if you set the prop to default to false in the component. This is generally better style.

PR #1348, opened 2022-10-19
vaultv1registry tests fixed

Amirjab21:


packages/hardhat/test/forge/fork/vault/VaultsV1Registry.t.sol
```diff
@@ -558,7 +571,7 @@ contract VaultsV1RegistryTest is Test {
     address vault = helper__deployVaultAndRegister(1, true);
 
     vm.expectEmit(false, false, false, true, address(vaultsV1Registry));
-    emit VaultStatusChanged(vault, true, false);
+    emit VaultStatusChanged(vault, false, false);
```
Amirjab21, line 574: Is this supposed to be false? We're not initialising this mapping value with anything so I assume it just takes on the false value.

PR #1393, opened 2022-11-03
Addressing Notes from VaultsV2 Review

K-G-J:
## Details
- Updated the Vault.sol contract to address notes discussed during the prior review
- Left comments with original notes from review in place to show where changes were made to address them
 
## After current changes are reviewed/approved: TODO
- Make an executive decision on the upper bound for fee structure (currently could be as high as 100%)
- Make an executive decision on the bounds for rage quitting after a new strategy is proposed 
- I believe the decimal math should be accurate, but tests need to be written/refactored for specific non-18-decimal cases to ensure
- Need to update tests to reflect changes with new ```propose change --> set change``` design for strategy and fee changing functions and permissions if implementation approved

### DIFF
--- a/packages/hardhat/contracts/core/defi/vault/Vault.sol
+++ b/packages/hardhat/contracts/core/defi/vault/Vault.sol
@@ -69,6 +69,7 @@ contract Vault is
   uint256 public proposalTimeStamp;
   uint256 public quitPeriod; // default is 3 days
   FeeStructure public proposedFees;
+  uint256 proposedFeeTimeStamp;
 
   //  EIP-2612 STORAGE
   uint256 internal INITIAL_CHAIN_ID;
@@ -567,6 +568,7 @@ contract Vault is
     ) revert InvalidFeeStructure();
 
     proposedFees = newFees;
+    proposedFeeTimeStamp = block.timestamp;
 
     emit NewFeesProposed(newFees);
   }
@@ -577,6 +579,9 @@ contract Vault is
   function setFees() external {
     // NOTE --> We can make this permissionless since it can only be changed to proposedFee
     // NOTE --> Add a proposal for fees (similar to changeStrategy) --> see proposeNewFees function above
+
+    // NOTE --> Brought in quitPeriod logic for rage quitting as in changeStrategy
+    if (block.timestamp < proposedFeeTimeStamp + quitPeriod) revert NotPassedQuitPeriod(quitPeriod);
 
     emit FeesUpdated(feeStructure, proposedFees);
     feeStructure = proposedFees;
--- a/packages/hardhat/contracts/core/defi/vault/Vault.sol
+++ b/packages/hardhat/contracts/core/defi/vault/Vault.sol
@@ -51,7 +51,7 @@ contract Vault is
 
   bytes32 public contractName;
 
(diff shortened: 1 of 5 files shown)


packages/hardhat/contracts/core/defi/vault/Vault.sol
```diff
@@ -470,16 +500,24 @@ contract Vault is
...
 
   /**
-   * @notice Propose a new strategy for this vault
+   * @notice Propose a new strategy for this vault. Caller must have VAULTS_CONTROlLER from ACLRegistry.
    * @param newStrategy A new ERC4626 that should be used as a yield strategy for this asset.
+   * @param _quitPeriod time for rage quit after new strategy is proposed, if not set defaults to 3 days
    * @dev The new strategy can be actived 3 Days after proposal. This allows user to rage quit.
    */
-  function proposeNewStrategy(IERC4626 newStrategy) external onlyRole(VAULTS_CONTROLLER) {
+  function proposeNewStrategy(IERC4626 newStrategy, uint256 _quitPeriod) external onlyRole(VAULTS_CONTROLLER) {
```
RedVeil, line 508: I think we should move the quitPeriod-setter into its own function since it also applies for fees. I wouldnt mix proposal and setter.

packages/hardhat/contracts/core/defi/vault/Vault.sol
```diff
@@ -221,10 +244,11 @@ contract Vault is
    *   HWM in a fee period, issue fee shares to the vault equal to the performance fee.
    */
   function accruedPerformanceFee() public view returns (uint256) {
-    uint256 shareValue = convertToAssets(1 ether); // NOTE --> check if we have to take decimals into account
+    uint256 shareValue = convertToAssets(1 ether);
 
     if (shareValue > vaultShareHWM) {
-      return feeStructure.performance.mulDivDown((shareValue - vaultShareHWM) * totalSupply(), 1e36); // NOTE --> Take decimals into account
+      return feeStructure.performance.mulDivDown((shareValue - vaultShareHWM) * totalSupply(), 1e36);
+      // NOTE --> Take decimals into account --> vaultShares will always be denominated in 18 decimals
```
RedVeil, line 251: Are you sure? Since we have the decimal overwrite that should be used. Do we have some tests confirming this?

RedVeil:
I already have the USDC Unit test which worked fine last time. I would like to have some additional test for the assumption of using 1e18 in conversions.

packages/hardhat/contracts/core/defi/vault/Vault.sol
K-G-J: added new separate setQuitPeriod function

packages/hardhat/contracts/core/defi/vault/Vault.sol
K-G-J: **correction**: vaultShares will always be denominated in decimals of **asset**
The USDC integration test does test this but it probably needs more robust testing

K-G-J:
> I already have the USDC Unit test which worked fine last time. I would like to have some additional test for the assumption of using 1e18 in conversions.

agreed, will begin this today

PR #1476, opened 2022-11-25
Test prettier action

Amirjab21:


Amirjab21:
This PR is good to go.

A thing to note:
1 - the rules set for format on save on my vs code editor seem to be different to the ones set on this CI formatting which is odd as I've copied the .prettierrc file and put it inside github workflows folder.

PR #1486, opened 2022-11-28
Feature/proxy component

SharifElfouly:


SharifElfouly:
- Implements the useMulticall hook
- the useMulticall uses the useProxy hook internally to get the correct proxy address if it exists
- Proxy.tsx is an example component that uses the useMulticall hook

